PR #1, opened 2026-05-16
Improve TargetBridge virtual display streaming

ghost:
## Summary
- capture the Sender's receiver-backed virtual iMac display instead of falling back to the MacBook desktop
- add lower-latency streaming behavior with reduced buffering, 30 fps 5K mode, higher 5K bitrate headroom, and backpressure frame dropping
- update Receiver fullscreen/display handling and cursor packet support for the full-screen iMac target workflow

## Verification
- git diff --check
- TargetBridge-Sender/scripts/build_targetbridge_sender_app.sh
- TargetBridge-Receiver/scripts/build_tbreceiver_c_app.sh

Note: Receiver build required the documented Homebrew dependencies: ffmpeg, sdl2, pkg-config, and dylibbundler.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/display.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/display.c
@@ -33,6 +33,9 @@ struct tb_display {
     int           quit;
     int           preferred_fullscreen;
     int           is_connected;
+    int           cursor_x, cursor_y;
+    int           cursor_source_w, cursor_source_h;
+    int           cursor_visible;
 
     char          last_ip[64];
     char          last_status[128];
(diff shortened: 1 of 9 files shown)


swellweb:
This is excellent work — the CGDisplayStream direct path, cursor overlay, TCP no-delay and backpressure are exactly the kind of improvements this project needed. The new Smooth/Smooth+/Crisp profiles make a lot of sense too. Merging this in, thank you!

swellweb:
Quick update: after merging I made a few follow-up changes on top of your work — fixed a virtual display creation failure caused by the new `activatePreferredMode` step timing out on some setups (made it non-blocking), and added a 'Large cursor' toggle in the sender UI that lets users switch between the native baked-in cursor (your SDL overlay hidden) and your SDL overlay cursor. The toggle is disabled while streaming so the setting takes effect on the next connection. Thanks again for the contribution!

PR #6, opened 2026-05-17
Memory leak fixes

ThomasWaldmann:


ThomasWaldmann:
Another few Gemini findings and fixes, see the extensive commit comments.

Please review critically, I can't and didn't review the changes.

swellweb:
Solid fixes — the retain cycle on stateUpdateHandler and the autoreleasepool in the mode polling loop are both real issues. The SCStream output removal before stopCapture is the right approach. Merging, thanks Thomas!

PR #15, opened 2026-05-18
README: avoid being overly specific if it works in general

ThomasWaldmann:


ThomasWaldmann:
The README often talked about a very specific usecase:
- an Apple Silicon MacBook
- an Intel iMac
- connected via TB3/TB4 cable

I think this might mislead readers so they think their usecase is not covered.

Guess it should work with any Mac that has Thunderbolt bridge and at least TB3, including Mac Minis (Intel + ARM), MacBooks (Intel + ARM), Mac Studio (ARM) as sender and also iMacs (Intel + ARM) as receiver.

The only part that (currently) needs to be specific is the release binaries offered.

Or am I missing something?

swellweb:
Applied the good parts directly in [2b0e815](https://github.com/swellweb/targetBridge/commit/2b0e815): removed "MacBook" specificity, generalized screenshot labels, updated download line formatting, added TB3/4/5 info to requirements, fixed "MacBook display" → "sender display".

Kept "Apple Silicon Mac" in requirements (sender binary is arm64-only and requires CGVirtualDisplay which is Apple Silicon + macOS 14+). Also kept "Intel iMac" on the receiver side since that is the only supported receiver hardware.

Closing in favor of the direct commit — thanks for the suggestions!

ThomasWaldmann:
But wouldn't it work with an Apple Silicon iMac also if one compiles the respective binary?

PR #18, opened 2026-05-18
Persist receiver IP in UserDefaults

ThomasWaldmann:


ThomasWaldmann:
Typed in (or copy&pasted) the same IP way too often while testing...

swellweb:
When you test the new branch, you will find the function to automatically look up the receiver's IP address in the sender's IP address.

PR #24, opened 2026-05-19
receiver: fix OOB reads in JSON packet parser

ThomasWaldmann:
The payload pointer handed to the packet callback pointed directly into the middle of the internal reassembly buffer.  No NUL byte was written after the payload bytes, so the string functions used in the three JSON field extractors in main.c (strstr, strchr, strtol) could read past the end of the packet into unrelated memory — or, when the payload was the last bytes of the buffer, into unmapped memory.

Fix: reserve one extra byte of capacity in tb_parser_feed and write a NUL sentinel immediately after each payload before invoking the callback. The sentinel byte is saved before the write and restored afterwards so that a following packet's length header is not corrupted when two complete packets are contiguous in the buffer.

No changes to main.c are required; the existing string functions now always see a NUL-terminated payload for the duration of the callback.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/net.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/net.c
@@ -49,7 +49,7 @@ static uint32_t read_be32(const uint8_t *p) {
 }
 
 int tb_parser_feed(struct tb_parser *p, const uint8_t *data, size_t n) {
-    if (parser_reserve(p, p->len + n) < 0) return -1;
+    if (parser_reserve(p, p->len + n + 1) < 0) return -1;  /* +1 for NUL sentinel */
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
This was discovered by Gemini and I let Claude Sonnet fix it (with a bit of interaction to get it nice).

I reviewed it and think the finding and the fix are both correct.

The fix is a bit of a hack as it temporarily patches a NUL byte behind the packet data that the callback will process. But guess this is safer and simpler than fixing all the str functions to their "n" counterparts in the callback function. Also, the latter could easily reintroduce the same issue if a non-"n" variant is added in future.

I thought about another option: adding the NUL byte after the JSON text in the sender. But that would still leave the receiver vulnerable to a malicious sender...

PR #29, opened 2026-05-20
Add multi-session direct receiver branch updates

swellweb:
## Summary

This PR brings `feature/multi-imac-direct` up to date with the stable `v1.3.x` line while preserving the new direct multi-iMac sender architecture.

## Included in this PR

- multi-session sender flow for direct multi-receiver Thunderbolt setups
- automatic Bonjour-based receiver discovery
- updated sender UI with a cleaner session-oriented layout
- integration of post-`v1.3.0` fixes into the branch
- repo-local build outputs under `build/`
- receiver build/tooling updates using `pkgconf`
- receiver minimum target lowered to macOS 11
- persistence of extended-display arrangement between reconnects
- safe partial integration of PR #16:
  - improved `CGDisplayMode` selection on the sender side
  - throttled cursor-only redraws on the receiver side

## Intentionally not included

The aggressive capture preset retuning proposed in PR #16 was left out on purpose to avoid regressions in the multi-session branch until it is benchmarked and reviewed more carefully.

## Verification

- sender rebuild succeeds from this branch
- receiver rebuild succeeds from this branch
- branch has been tested locally and is currently behaving correctly

## Notes

This PR is focused on the direct multi-iMac branch only.
Relay mode (`MacBook -> iMac1 -> iMac2`) remains a separate future branch by design.


swellweb:
After merged this PR on main I will create the new release TargetBridge 2.0, Are you agree @ThomasWaldmann ?

ThomasWaldmann:
Yay, it works! Nice UI redesign!

2x iMac 5K (2017+2020) each directly connected via TB to MacBook.
Each 5K display is playing a youtube video at 15 FPS.

ThomasWaldmann:
After you merged into main, I will remove the cursor and mode parts of PR #16 and rebase it onto current main branch then.

README.md
```diff
@@ -10,29 +10,48 @@ If TargetBridge is useful to you, a ⭐ on GitHub helps others find it.
...
+This branch, `feature/multi-imac-direct`, contains additional in-progress work that is not part of the latest release yet:
+
+- direct multi-iMac sender support with multiple simultaneous sessions
+- automatic receiver discovery via Bonjour
+- remembered extended-display arrangement for the same receiver
+- repo-local build outputs under `build/`
+- receiver build target lowered to **macOS 11+** for Intel testing
+
+If you just want the stable public version, use the latest release. If you want to test the upcoming direct multi-iMac work, build from this branch.
+
```
ThomasWaldmann: guess this will need an update when it gets merged into main.

README.md
```diff
@@ -10,29 +10,48 @@ If TargetBridge is useful to you, a ⭐ on GitHub helps others find it.
...
 
 **[→ Download latest release (pre-built apps, no Xcode needed)](https://github.com/swellweb/targetBridge/releases/latest)**
 
 - `TargetBridge.app.zip` — Sender (Apple Silicon Mac, **requires macOS 14 Sonoma or later**)
-- `TargetBridge-Receiver.app.zip` — Receiver (Intel iMac, **requires macOS 13 Ventura or later**)
+- `TargetBridge-Receiver.app.zip` — Receiver (Intel iMac, **requires macOS 11 Big Sur or later**)
 
 Unzip and double-click. On first launch, grant Screen Recording permission to the sender.
 
+If you are testing the current development branch instead of the latest release, build outputs now go into:
```
ThomasWaldmann: Same here.

README.md
```diff
@@ -57,9 +76,26 @@ Unzip and double-click. On first launch, grant Screen Recording permission to th
...
+```
+
+Current branch behavior:
+
+- one sender app can manage multiple receiver sessions
+- each session can bind to its own Thunderbolt Bridge interface
+- discovered receivers can be selected from the UI instead of typing the IP manually
+- the main target is `extended + extended`; multi-session mirror mode still needs more testing
+
+This is branch-only work for now and should be considered preview / test functionality until it lands in a stable release.
```
ThomasWaldmann: And here.

README.md
```diff
@@ -70,5 +106,7 @@ If the receiver does not fill the iMac panel or the cursor/desktop feels scaled
 
 ## Quick start
 
-- Italian: `TargetBridge-QuickStart-IT.md`
-- English: `TargetBridge-QuickStart-EN.md`
+- Italian: `docs/QuickStart-IT.md`
+- English: `docs/QuickStart-EN.md`
+
+When testing this branch, the sender can also discover compatible receivers automatically in the UI and prefill their Thunderbolt Bridge IP.
```
ThomasWaldmann: And here.

ThomasWaldmann:
"persistence of extended-display arrangement between reconnects"

Guess that does not work. I used extended displays and positioned them (via macOS Display settings) above left and above right of my MB display.

Disconnected all sessions.

Reconnected.

Now both iMac displays are to the right of the MB display.

I can reproduce.

swellweb:
I pushed a follow-up fix for the extended-layout persistence issue.

Probable causes were:
- identical 5K receivers sharing the same persistence key
- `stopAll()` disconnecting sessions sequentially, which could let macOS rearrange the remaining displays before their positions were saved

The update now:
- uses a receiver-specific persistence key
- snapshots all extended-display arrangements before `stopAll()` tears the sessions down

Commit: `adac7c6`

Could you please retry the same reconnect scenario with the updated branch and see whether the two extended displays now come back to their saved positions?

ThomasWaldmann:
It still behaves the same, bug persists.

ThomasWaldmann:
I can also reproduce this with only a single display, it does not remember that my macbook shall be arranged "below".

swellweb:
I pushed a follow-up fix for the remaining extended-layout persistence issue.

Root cause: we were still persisting the virtual display position as absolute global `x/y` coordinates. That works if the main display stays at the expected origin, but it loses the intended relationship when the MacBook should be positioned below or otherwise moved relative to the virtual display.

The update now stores the extended-display arrangement as an offset relative to the main display (`dx/dy`) and reapplies that relative placement on reconnect.

Commit: `38b990d`

One note for retesting: existing saved values only contain the old absolute `x/y` format, so the first reconnect after updating may still reuse the old placement once. After moving the display again with the updated build and disconnecting, subsequent reconnects should use the new relative persistence format.

Could you please retry the single-display case where the MacBook should stay below and see whether it now comes back correctly after one save/reconnect cycle?

ThomasWaldmann:
Still doesn't remember the arrangement. After stopping the connections, reconnecting arranges the iMac displays always to the right of the MB display.

That was for 2 displays, now retrying just with 1.

With 1 iMac: same thing, the iMac display (which i positioned above the MB display) always resets to be to the right.

swellweb:
I found the remaining issue: the saved `dx/dy` values were already relative to the main display, but the restore path was still applying them as if they were absolute global coordinates.

That meant layouts like "iMac above" or "MacBook below" would fall back to the default placement on reconnect.

Fixed in `52a729a`:
- `dx/dy` entries are now restored relative to the current main-display origin
- legacy `x/y` entries are still treated as absolute coordinates for backward compatibility

This one should not require re-saving the layout first if the defaults already contain `dx/dy` from a previous run. Could you retry with the latest branch build?

ThomasWaldmann:
Nope, no improvement.

swellweb:
I re-tested the latest branch build locally and can now confirm the extended-layout persistence works on my side.

Test sequence:
1. Connected a single iMac in extended mode
2. Moved the iMac display to the left of the MacBook in macOS Display Arrangement
3. Stopped TargetBridge completely
4. Started the connection again
5. The display position remained unchanged after reconnect

So at least in this retest, the layout no longer snapped back to the right-hand default placement.

I also captured before/after screenshots during the test sequence on my side.

ThomasWaldmann:
Can you review this (made by Gemini), this made it work for me:
```
diff --git a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
index e25f0cd..2dea14e 100644
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
@@ -1123,6 +1123,8 @@ final class TBDisplaySenderSession: NSObject, ObservableObject, Identifiable, @u
         Task { @MainActor [weak self] in
             guard let self else { return }
 
+            var hasAppliedArrangement = false
+
             for attempt in 1...12 {
                 try? await Task.sleep(nanoseconds: 500_000_000)
 
@@ -1131,12 +1133,17 @@ final class TBDisplaySenderSession: NSObject, ObservableObject, Identifiable, @u
                       self.activeProfile != nil
                 else { return }
 
-                if CGDisplayIsInMirrorSet(virtualDisplayID) == 0 {
+                // Even if not in a mirror set initially, we must configure the extended desktop
+                // at least once to apply our custom saved/default layout arrangement.
+                if CGDisplayIsInMirrorSet(virtualDisplayID) == 0 && hasAppliedArrangement {
                     self.displayStateText = self.describeDisplayState(for: virtualDisplayID)
                     return
                 }
 
                 let configured = self.configureExtendedDesktop(for: virtualDisplayID)
+                if configured {
+                    hasAppliedArrangement = true
+                }
                 self.displayStateText = self.describeDisplayState(for: virtualDisplayID)
                 NSLog(
                     "TargetBridge: extended desktop recovery attempt %d for %u configured=%d state=%@",
@@ -1146,7 +1153,7 @@ final class TBDisplaySenderSession: NSObject, ObservableObject, Identifiable, @u
                     self.displayStateText
                 )
 
-                if configured || CGDisplayIsInMirrorSet(virtualDisplayID) == 0 {
+                if configured || (CGDisplayIsInMirrorSet(virtualDisplayID) == 0 && hasAppliedArrangement) {
                     return
                 }
             }
```
Explanation by Gemini:
```
fix(sender): apply saved extended layout on reconnection

Ensure that `configureExtendedDesktop` is executed at least once during the
extended desktop recovery process, even if the virtual display is not initially
in a mirror set.

Since virtual display sessions generate randomized product IDs and serial numbers
to avoid collisions, macOS treats each reconnection as a brand-new display and
defaults to placing it to the right of the internal display without mirroring.

Previously, the recovery loop would immediately return early upon seeing
`CGDisplayIsInMirrorSet == 0` on its first check, completely bypassing
`configureExtendedDesktop` and preventing the custom saved arrangement
(e.g., placing the virtual display ABOVE the built-in display) from being restored.

We now track whether the layout configuration has been successfully applied
(`hasAppliedArrangement`) and only allow the early exit after a successful
configuration pass has been completed.
```

ThomasWaldmann:
In my tries, I did not stop TB completely, I just disconnected and reconnected the displays via the buttons.

The code in the diff fixed that. And it also worked when completely terminating the app, which I tried afterwards.

swellweb:
Adding the two screenshots from my local retest for reference:

- Session live in `Extended Desktop` mode before stopping TargetBridge: [layout-persistence-session-live.png](https://github.com/swellweb/targetBridge/blob/ce06d28754e6570e509e6372d4348110d8311c02/images/layout-persistence-session-live.png)
- Display Settings state after reconnect, with the position preserved: [layout-persistence-after-reconnect.png](https://github.com/swellweb/targetBridge/blob/ce06d28754e6570e509e6372d4348110d8311c02/images/layout-persistence-after-reconnect.png)

Test sequence on my side:
1. I first positioned the iMac display to the left of the MacBook
2. I stopped TargetBridge completely
3. I started the connection again
4. The display position stayed unchanged after reconnect

ThomasWaldmann:
@swellweb Tried that here and it was reset to right-of-internal (using 52a729a559737299edf54a2af0869e57dea253a3).

With code from diff, it worked.

ThomasWaldmann:
718a85f works for me!

PR #34, opened 2026-05-20
docs: hardware notes about Macs, TB cables and adaptors

ThomasWaldmann:


ThomasWaldmann:
@swellweb If you like, quite a bit of hardware related infos could be removed from the README and it could just refer to the hardware docs.

swellweb:
thank you for your support.

PR #36, opened 2026-05-20
Add security policy for supported versions and reporting

ThomasWaldmann:


ThomasWaldmann:
@swellweb A simple policy, just as a suggestion.

That reddit bot always gives this repo a warning because it has no secpol.

Also, it is a useful standard for security researchers.

swellweb:
wow, this is what I needed. :)

PR #46, opened 2026-05-21
implement scale-independent cursor shape synchronization, fixes #45

### LINKED ISSUE #45: cursor does not change shape on receiver-side display
Usually the cursor changes shape at window borders or corners when it is in the right position to resize the window, but here it just stays in the normal shape, making you guess.

I activated "big cursor" setting, not sure if related.

ThomasWaldmann:
- Implement 32x32 transparent canvas normalization in TBDisplaySenderService.swift to flatten normal vs. HiDPI scaling differences

- Dynamic signature registration of standard and private Cocoa resize cursors

- Real-time cursor shape detection mapped against normalized signatures

- Low-latency vector shape drawing in display.c for vertical, horizontal, NWSE diagonal, and NESW diagonal double-arrows

- Update protocol to serialize and transmit cursor shape types alongside coordinates

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/display.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/display.c
@@ -36,6 +36,7 @@ struct tb_display {
     int           cursor_x, cursor_y;
     int           cursor_source_w, cursor_source_h;
     int           cursor_visible;
+    int           cursor_type;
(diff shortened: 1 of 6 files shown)


ThomasWaldmann:
Gemini FTW! :-)

It works, tested it manually.

PR #51, opened 2026-05-22
feat(i18n): add Chinese (zh-Hans) UI and Settings sheet refactor

Yananix:
## Summary

Adds Simplified Chinese (zh-Hans) UI to both the macOS sender app and the C receiver, and refactors sender preferences into a Settings sheet for a cleaner main panel.

Default UI language is now auto-detected from the system locale (`Locale.preferredLanguages` on the sender, `LC_ALL` / `LC_MESSAGES` / `LANG` plus `TBR_LANG` override on the receiver). English / Italian / German behavior is preserved.

## Sender (SwiftUI)

- `TBDisplaySenderLocalization.swift`
  - Add `.chinese` case to `TBDisplaySenderLanguage`, picker title `中文`.
  - Auto-detect default language from `Locale.preferredLanguages` (zh / it / de / en prefix match), fall back to English.
  - Translate ~80 user-facing strings into Chinese.
  - New strings for the Settings sheet (title, Done button, language / appearance / about section headers, gear button accessibility label).
- `TBDisplaySenderContentView.swift`
  - Remove the standalone `languageCard` and `settingsCard` from the main panel.
  - Add a gear button at the top-right of the header card; tapping opens a new `TBDisplaySenderSettingsView` sheet (460 × 420) with three sections: Language, Appearance, About.
  - Add `.chinese` branches to all helper titles (routing / output / session monitor / live / connected / idle).
- `TBDisplaySenderService.swift`
  - `TBDisplayCaptureSource.title` adds `.chinese` cases (`镜像桌面`, `扩展桌面`).

## Receiver (C / SDL2 + CoreText)

- New `src/i18n.h` — header-only inline module, no Makefile changes.
  - `tb_locale_is_chinese()` with cached lookup, priority `TBR_LANG` > `LC_ALL` > `LC_MESSAGES` > `LANG`.
  - `tb_tr(en, zh)` macro-style helper for inline string selection.
- `src/display.c`
  - Switch banner fonts based on locale: `PingFangSC-Semibold` / `PingFangSC-Regular` for Chinese (CoreText cannot render CJK with Helvetica/Menlo), `Helvetica` family for ASCII.
  - Translate all 11 banner strings (title, subtitle, three column headers, display / stream profile labels, three-line footer).
- `src/main.c`
  - Translate all dynamic status / sender / display / mode / panel strings (`stream active`, `sender connected`, `waiting for sender`, etc.).
  - Fix `bonjour_update()` literal-string guard to recognize both `not detected` and `未检测到`.

## How to test

```sh
# Receiver (auto, follows LANG / LC_ALL)
./tbreceiver --windowed

# Force Chinese banners
TBR_LANG=zh ./tbreceiver --windowed

# Force English banners
TBR_LANG=en ./tbreceiver --windowed
```

For the sender, set system language to Simplified Chinese (or click the gear → Language → 中文).

## Verification

- ✅ Sender Swift type-check passes (`xcrun swiftc -typecheck`), private `CGVirtualDisplay*` API is provided by the bridging header as before.
- ✅ Sender full Xcode build succeeds (Xcode 26.5, macOS 26.5 SDK), `.app` runs and shows Chinese UI on a `zh-Hans-CN` system.
- ✅ Receiver compiles cleanly with `make` (Apple Silicon, brew ffmpeg / sdl2 / pkgconf).
- ✅ `i18n.h` runtime fallback verified across 5 scenarios (default, `TBR_LANG=zh`, `TBR_LANG=en` overriding `LANG=zh_CN`, `LANG=zh_CN`, `LANG=en_US`).
- ✅ No functional changes — existing toggles (`showsMenuBarIcon`, `largeCursor`) preserved, just relocated from the old in-panel `settingsCard` into the new gear sheet.

## Notes

- Chinese is opt-in by system locale; existing English / Italian / German behavior is unchanged for users who do not run in `zh-*`.
- The receiver gracefully falls back to English banners if the Chinese fonts (`PingFangSC-*`, ships with macOS) are absent.

ScreenShot_2026-05-22_210730_804

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/display.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/display.c
@@ -6,6 +6,7 @@
  */
 
 #include "display.h"
+#include "i18n.h"
 
 #include <CoreFoundation/CoreFoundation.h>
 #include <CoreGraphics/CoreGraphics.h>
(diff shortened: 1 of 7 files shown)


Yananix:
cc @ThomasWaldmann

swellweb:
Thanks a lot for this — the Chinese work is definitely valuable, and I really appreciate the effort.

The only reason I would hesitate to merge this PR as-is is that we are currently moving the project toward a shared JSON-based localization system for both Sender and Receiver, instead of keeping translations hardcoded in Swift/C.

So the direction is very similar, but the integration path is now a bit different from the one in this PR.

That said, I would absolutely like to use the Chinese translation work here.
If you’re open to it, I’d be happy to adapt/reuse your Chinese strings as the basis for a new `zh.json` language file in the shared localization system, rather than losing the work.

The Settings sheet refactor also overlaps with some UI changes already happening on our side, so I’d probably separate the “Chinese translation” part from the “UI/settings refactor” part when integrating it.

So in short:
- the PR is appreciated
- the Chinese translation is useful and we want it
- but we’ll likely integrate it through the new shared JSON localization path

Yananix:
@swellweb No problem, hope this helps.

ThomasWaldmann:
@swellweb You could also merge this to main.

If you want to move the stuff to json, you can move it with zh_CN then.

Yananix:
I've already extracted every string from this branch into an i18n-style JSON: `localization/translations.draft.json` (commit e8de124). 
Feel free to reuse it as the source for `zh.json` and drop the Swift/C edits. Marking the PR as draft.

PR #53, opened 2026-05-22
Fix virtual display capture in desktop mirror mode, fixes #52

### LINKED ISSUE #52: Error connecting with Duplicate Desktop: no virtual SCDisplay available (target=4, baseline=[3], available= [3], online= [3, 4])
Unable to connect to my iMac from two different machines, same error:
nessun SCDisplay virtuale disponibile (target=28, baseline=[1], disponibili=[1], online=[1, 28])
(except the target value, that is different between the machines)

Extended Display works and I can even manually change the Display settings to make the remote monitor the main, but I can't connect with Duplicate.

Using the v2.0 release downloads (First time user; never installed the earlier builds so unable to say whether this would've worked before.)

Senders:
Macbook Air M2 running Tahoe, 26.5
Mac mini M4 running Tahoe, 26.5

Receiver:
iMac 2017 running Sonoma 14.8.7 with OCLP

On both Senders, I'm able to see the Mirror in System Settings\Displays but then it eventually disappears, and then I see that error in the client.
On the Receiver, I see the Sender connected and "profile sent" before it produces that error.

Stuff I've done:
For the Sender machines, I had an error running it and used the xattr command. I've given the app the required permissions and rebooted a bunch of times. I installed homebrew on the receiver and ffmpeg and sdl2 as suggested on the reddit thread.

Tried with two different cables as well, both test above 10GBPS. Tried different TB ports on all the machines.

Any suggestions on what I might be doing wrong?

(As an aside, I use Hammerspoon to automate tasks on my machine. If there was a CLI version of this, I could likely just automate switching the display settings after Extend and that would be good enough.)

ThomasWaldmann:
In duplicate/mirror desktop mode, the virtual display created by TargetBridge is configured to mirror the main display.

Because ScreenCaptureKit (`SCShareableContent`) does not include mirrored displays in its shareable display list, attempting to wait for and capture the virtual display ID resulted in a timeout and threw the 'no virtual SCDisplay available' error.

This PR resolves this by capturing the main display (`CGMainDisplayID()`) directly instead of the virtual mirror display when `captureSource` is `.desktopMirror`.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
@@ -1000,8 +1000,9 @@ final class TBDisplaySenderSession: NSObject, ObservableObject, Identifiable, @u
         do {
             let preset = capturePreset
 
-            if captureSource == .extendedDesktop, session.displayID != kCGNullDirectDisplay {
-                if startDirectDisplayStream(displayID: session.displayID, preset: preset) {
+            if session.displayID != kCGNullDirectDisplay {
+                let captureDisplayID = (captureSource == .desktopMirror) ? CGMainDisplayID() : session.displayID
+                if startDirectDisplayStream(displayID: captureDisplayID, preset: preset) {
                     return true
                 }
             }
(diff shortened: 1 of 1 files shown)


swellweb:
fix in new release with netlink.

PR #54, opened 2026-05-22
fix: Update timestamp of built TargetBridge.app bundle

ThomasWaldmann:
During incremental builds, xcodebuild only updates files inside the bundle (under Contents/MacOS) without updating the modification timestamp (mtime) of the top-level .app directory in Xcode's DerivedData folder. When the build script copies the bundle via ditto, ditto preserves these timestamps exactly as they are in the source, resulting in Finder showing a stale build date (e.g. from days ago).

Adding a touch on the destination app bundle ensures the directory modification time is updated to the current build time, displaying the correct build date in Finder.


ThomasWaldmann:
Without this "touch", it is quite annoying: if you look into `build/` after having built something new, you are always wondering whether it actually worked, because it is showing some timestamp for 3 days ago or so.

swellweb:
It's okay thank you!

PR #55, opened 2026-05-23
Fix resource leaks

ThomasWaldmann:


ThomasWaldmann:
I just told Gemini "fix resource/memory leaks" multiple times, this is the result for review.

PR #12, opened 2026-05-18
Force OpenGL renderer backend to resolve macOS Tahoe Metal flickering, fixes #11

ThomasWaldmann:
Problem Diagnosis:
Under macOS Tahoe, SDL2's default Metal renderer backend suffers from severe swapchain presentation and compositor synchronization conflicts:
Variable frame-presentation rates (caused by high-rate cursor updates interleaved with slower video frames) trigger Apple's dynamic WindowServer compositor adjustments (such as EDR/HDR, ProMotion, and TrueTone transitions), causing the entire screen to constantly flicker.

How this change helps:
Forces the renderer to utilize the OpenGL backend driver via SDL_HINT_RENDER_DRIVER. OpenGL is implemented via Apple's mature, ultra-stable OpenGL-to-Metal translation stack. It utilizes a rock-solid, stable frame presentation swapchain that completely eliminates visual flickering, while maintaining full GPU hardware acceleration.

Issue #11, opened 2026-05-18
Whole screen is flickering all the time


ThomasWaldmann:
Got the TB5 cable and connected the MBP to the iMac.

It basically works, but is not really usable, because the whole screen is flickering all the time, e.g.:
- when i move the mouse cursor
- when i type something
- even when doing nothing

Flickering means that the colour values are jumping up and down, e.g. the black toggles between really black and dark grey. This happens for all colours.

- MBP: MacBook Pro M3 Pro, macOS Sequoia 15.7.7
- iMac: Intel iMac 5k 27" (2020), macOS Tahoe 26.5

It does not depend on the stream resolution I am selecting in the sender app, happens in all resolutions.

Sender-FPS with 5K 5120x2880@48 HEVC stream is between 2 and 16 FPS.

ThomasWaldmann:
I could fix this with the help of Gemini, see #12.

PR #56, opened 2026-05-23
Prefer GPU-backed receiver rendering

DrDavidL:
## Summary
- prefer SDL's Metal renderer on macOS, with OpenGL/default accelerated fallback
- allow TB_RECEIVER_RENDER_DRIVER to force a renderer for troubleshooting
- prevent the active receiver loop from busy-spinning when the nonblocking socket has no data

## Validation
- Built receiver bundle locally with PATH=/opt/homebrew/bin:/usr/local/bin:$PATH TargetBridge-Receiver/scripts/build_tbreceiver_c_app.sh
- Local bundle: build/TargetBridge Receiver.app

## Notes
- Local bundle was built on Apple Silicon and is arm64. Needs Intel iMac build/test before marking ready.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/display.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/display.c
@@ -1,6 +1,7 @@
 /* display.c — SDL2 NV12 renderer.
  *
- * SDL2 on macOS uses Metal renderer by default (SDL_RENDERER_ACCELERATED).
+ * On macOS, prefer SDL's Metal renderer and fall back to OpenGL/default
+ * accelerated backends if needed.
  * SDL_PIXELFORMAT_NV12 + SDL_UpdateNVTexture lets us upload YUV planes
  * directly to GPU; the shader does YUV→RGB conversion on the GPU.
  */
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
What about #11 and #12?

I tested the code (using defaults, no env var set):

Both: connected via TB cable, same receiver binary (code from this PR), extended display at 5K.

## Intel iMac 5K 27" 2020, Tahoe 26.5
Colours are off, flickers all the time (especially, but not only, when moving the mouse cursor). This PR would basically revert my fix from #12 "by default".

## Intel iMac 5K 27" 2017, Ventura 13.7.8
Interesting: I don't see flickering there, colours are ok.

So guess it is either a hw bug in the 2020 model or a Tahoe bug.

ThomasWaldmann:
The busy-spin fix seems to be out of scope of this PR, maybe it would be simpler to have it in a separate PR.

swellweb:
I think it's bettere separate the fix in another branch but not for this version but for the coming soon versione with netlink.

DrDavidL:
No problem. Thanks for testing. Works fine on my 2019 iMac! I'll try to keep pace with main updates for future consideration.

DrDavidL:
Rebased on latest main (b916f1d → bb457f3). Standing by for the additional update you mentioned before merging.

swellweb:
Thanks again for this PR — I've now pulled the safe parts of it into our current `feature/languages` branch.

What I kept:
- renderer override support via `TB_RECEIVER_RENDER_DRIVER`
- a more robust accelerated renderer selection path
- the socket busy-spin fix / idle yield improvement in the receiver loop

What I intentionally did *not* keep as-is is preferring Metal by default on macOS.
Thomas' testing on the 2020 5K iMac + Tahoe 26.5 showed flicker / wrong colours there, so I kept the default order conservative:
- `opengl` first
- `metal` as fallback
- explicit override still possible via `TB_RECEIVER_RENDER_DRIVER=metal`

So the idea from this PR is definitely useful, but I'm integrating it in a way that preserves the safer Tahoe behaviour by default.

If you want, we can later still split the renderer/backend tuning into a smaller dedicated PR once we have more hardware-specific data.

DrDavidL:
Closing this as **superseded by `main`**. Both changes here have already landed (in equivalent or deliberately-adjusted form):

- **Renderer selection** — `main` now has the same infrastructure this PR introduced: a `tb_disp_try_renderer` helper, the `TB_RECEIVER_RENDER_DRIVER` override, and an accelerated-renderer fallback list (`display.c`). However, `main` orders the macOS drivers **`{ "opengl", "metal" }`** (OpenGL first) — intentionally the *opposite* of this PR's `{ "metal", "opengl" }`. That was a deliberate choice to avoid Metal's HDR/swapchain flickering on macOS Tahoe, so re-preferring Metal here would be a regression.
- **Busy-spin fix** — the `socket_activity == 0` yield guard in the main loop is already present on `main`.

No remaining unique value, so closing. If the Metal-vs-OpenGL ordering is worth revisiting (e.g. it's better on Apple Silicon while worse on Tahoe), that's best raised as a focused issue/PR on `main` rather than reviving this one.

PR #58, opened 2026-05-23
update and restructure README, fixes #57

### LINKED ISSUE #57: README is outdated
Although there is 2.0.1 release, the readme still says "The latest packaged public release is still v1.3.0."

Suggestion:

Let's not talk about releases in the readme.

If we want something more than the per-release infos in github releases, there could be `docs/CHANGES.md` with the change log / history.

To avoid conflicts in there, updates should be made after merging something into main (or a maintenance branch, in case we'll have such in the future).

ThomasWaldmann:
see commit comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,27 +2,65 @@
 
 # TargetBridge
 
+Apple dropped Target Display Mode in late 2014 with the 5K iMac — and it never came back.
+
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Best is to click on `... -> view file` rather than reading the diff.

swellweb:
Sorry but today is a very a mess...

swellweb:
Thank you for your PR.

PR #69, opened 2026-05-26
Sender: GUI options for screensaver/wake handling, stable extended-display identity, diagnostic logging

DrDavidL:
## Summary

Sender-side GUI options that address two separate observed issues — windows on the extended display being forgotten after each reconnect, and intermittent post-event stutter — plus diagnostic plumbing for a longer-duration stutter that is still being characterized.

After in-field testing of the original screensaver-focused commit, both screensaver-related toggles default to **off** in this update — they didn't fix the long-duration stutter reproduced by the user and the actual repro cause appears to be unrelated to display sleep. They remain available as options for setups where they help.

## Changes

**Stable extended-display identity** (`ReceiverBackedVirtualDisplaySession.swift`)
- The extended-desktop virtual display previously generated a random productID/serialNumber on every `extendedDesktop()` call. macOS keys window-placement memory on those fields, so every reconnect looked like a brand-new display and windows on it were not restored.
- Identity is now derived deterministically (djb2) from `profile.receiverName + panelWidth + panelHeight` — same fields the codebase already uses for the persisted extended-arrangement key. Same receiver → same virtual display identity across reconnects → window placement persists.
- `TBDisplayCaptureSource.virtualDisplayIdentity` now takes a `TBMonitorDisplayProfile`.

**Settings card toggles** (`TBDisplaySenderContentView.swift`, `TBDisplaySenderManager.swift`, `TBDisplaySenderService.swift`)
1. *Prevent screensaver / display sleep while streaming* — default **off**. When on, adds `.idleDisplaySleepDisabled` to `ProcessInfo.beginActivity` so the screensaver cannot engage during a session.
2. *Auto-restart capture after wake / unlock* — default **off**. When on, observes `screensDidWakeNotification`, `com.apple.screenIsUnlocked`, and `com.apple.screensaver.didstop`, and soft-restarts the capture pipeline (`SCStream`/`CGDisplayStream` + `VTCompressionSession` + capture timers + activity) while preserving the `NWConnection` and the virtual display. PTS sequence is reset to avoid receiver-side pacing desync.
3. *Log virtual display events to Console (verbose)* — default **off**. When on, registers `CGDisplayRegisterReconfigurationCallback` to log every add/remove/mirror/mode-change event affecting any display, and emits a 60 s stream snapshot to `NSLog` containing streaming state, FPS, virtual display online status, pending packet count, in-flight encode count, and PTS sequence. Intended for diagnosing the long-duration stutter from `Console.app` without leaving heavy logging on for ordinary users.

**Per-session button**
- *Restart capture* (enabled while streaming) — runs the same soft-restart path on demand. Notably this does **not** rebuild the virtual display or `NWConnection`; the long-duration stutter so far does not appear to be cleared by this and remains under investigation (see the verbose logging toggle).

UserDefaults keys: `fd.tbdisplaysender.preventDisplaySleep`, `fd.tbdisplaysender.autoRestartOnWake`, `fd.tbdisplaysender.verboseDisplayLogging`. Existing stored values are preserved across the default flip.

Localized in IT / EN / DE.

## Test plan

- [ ] Build via Xcode (`TBDisplaySender` scheme) — verified locally, builds clean.
- [ ] Default install (all three toggles off) behaves identically to current `main`.
- [ ] With extended-desktop capture mode and window placement set up on a receiver, disconnect → reconnect — windows return to the extended display rather than reflowing to the laptop. Test with at least two distinct receivers to confirm identities don't collide.
- [ ] Enable "Prevent screensaver / display sleep" + let the Mac idle past the screensaver timeout while streaming — screensaver should not engage.
- [ ] Enable "Auto-restart capture after wake" only + manually trigger a screensaver and dismiss it — soft-restart logs should fire (`TargetBridge: system wake — soft restart of capture pipeline`).
- [ ] Click "Restart capture" while streaming — capture briefly drops and reestablishes; `NWConnection` and virtual display remain.
- [ ] Enable "Log virtual display events to Console (verbose)" and stream — `Console.app` shows reconfiguration events and the 60 s snapshot lines.
- [ ] All three settings persist across app relaunch.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -18,14 +18,27 @@ struct TBVirtualDisplayIdentity {
         usesDedicatedArrangementIdentity: false
     )
 
-    static func extendedDesktop() -> TBVirtualDisplayIdentity {
-        let random = UInt32.random(in: 0x0100...0xFFFE)
+    static func extendedDesktop(for profile: TBMonitorDisplayProfile) -> TBVirtualDisplayIdentity {
+        // Deterministic identity per receiver so macOS retains window placement
+        // and the saved extended-desktop arrangement across reconnects.
+        let key = "\(profile.receiverName)|\(profile.panelWidth)x\(profile.panelHeight)"
+        let hash = djb2(key)
+        let productLow = (hash & 0x00FF) | 0x01
+        let serialLow = (hash & 0xFFFE) | 0x0100
         return TBVirtualDisplayIdentity(
-            productID: 0x6000 | (random & 0x00FF),
-            serialNumber: 0x2027_0000 | random,
+            productID: 0x6000 | productLow,
+            serialNumber: 0x2027_0000 | UInt32(serialLow),
             displayNamePrefix: "TB Extend",
             usesDedicatedArrangementIdentity: true
         )
+    }
+
+    private static func djb2(_ input: String) -> UInt32 {
+        var hash: UInt32 = 5381
+        for byte in input.utf8 {
+            hash = hash &* 33 &+ UInt32(byte)
+        }
+        return hash
     }
 }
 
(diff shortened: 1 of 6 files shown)


swellweb:
Thank @DrDavidL  I put the Pr in the new release.

DrDavidL:
Superseded by #82, which contains these commits (`4f6fd8f`, `46b2c79`) plus the rest of the long-session-stutter fix. Collapsing the stack into #82 so it lands as a single PR.

PR #79, opened 2026-05-27
fix(sender): add connection timeout to prevent hanging on bad receiver IP

### LINKED ISSUE #76: add timeout when trying to connect to a receiver
When the receiver IP is bad (non-existing IP), it tries to connect... for rather long...

Not sure how long, quitting the sender app was faster than waiting.

I don't think it should wait longer than a few seconds, we can't work with super slow connections anyway.

ThomasWaldmann:
fixes #76


swellweb:
It's Fantastic Thank you @ThomasWaldmann

PR #81, opened 2026-05-28
Sender: capture watchdog + UI re-render storm mitigations

DrDavidL:
## Summary

Addresses the long-duration stutter that has plagued long-running sender sessions. **Stacked on #69** — the watchdog reuses the `softRestartCapture(for:)` path introduced there. If #69 lands first, this PR can be rebased onto main with no conflicts; if this lands first, please merge or rebase #69 onto it.

## What we learned from a 23.5 h live capture

After collecting `sample`, `vmmap`, `nettop`, and 24 h of `log show` from a sender process that was stuck at 99.7% CPU (write-up in user's local `~/tb-diag-20260527-203102/FINDINGS.md`):

- **`replayd` (the daemon backing `SCStream` / `ScreenCaptureKit`) was issuing `RPDaemonProxy: connection INTERRUPTED` every ~30 minutes** — 14 times in the observed window. This is the sandbox-extension refresh cycle. The interruption does **not** reliably reach our `SCStreamDelegate didStopWithError` handler, so the existing pipeline kept running in a degraded state.
- Sample showed the main thread spinning in recursive `NSView _layoutSubtreeWithOldSize:` — a SwiftUI/AppKit runaway-layout loop, almost certainly amplified by the degraded capture state.
- The virtual display was still online — earlier hypotheses about display eviction were wrong.
- The `NWConnection` had accumulated 2.1 M retransmits before becoming unresponsive, but that was downstream of the wedged main thread, not the cause.
- At startup SwiftUI logged "Publishing changes from within view updates is not allowed" 16 times — a latent bug that is harmless under normal load but amplifies the failure mode once anything degrades.

### 1. Capture watchdog (`TBDisplaySenderService.swift`)

- New `private var lastCaptureFrameAt: Date` is bumped at the top of both `encode(_ sampleBuffer:)` and `encodeDisplaySurface(_:displayTime:)`.
- New `captureHealthWatchdog: Timer` fires every 5 s on the main actor. If `isStreaming` is true, `activeProfile` is set, no restart is already in flight, and `Date().timeIntervalSince(lastCaptureFrameAt) >= 8.0`, the watchdog calls `scheduleCaptureRestart(reason: "watchdog (...)", delaySeconds: 0.5)`.
- The restart path is the **same `softRestartCapture(for:)` from #69** — capture pipeline, encoder, FPS / cursor timers, and the streaming activity are torn down; `NWConnection` and the virtual display are preserved.
- Started after `startFPSTimer()` in both `startCapture(for:)` and `startDirectDisplayStream(...)`. Torn down in `stop(resetStatusTo:)` and at the top of `softRestartCapture(for:)`.

Threshold rationale: 8 s is comfortably longer than any expected per-frame gap at 30 / 60 fps (33 / 16 ms) but short enough that the user only sees a brief blink before the rebuild completes.

### 2. Remove redundant `objectWillChange.send()` in `TBDisplaySenderManager.swift`

`@Published` already emits `objectWillChange` automatically. The didSet blocks, the `discoveryCancellable` sink, `addSession`, `removeSession`, `refreshBridgeInterfaces`, and `applyDiscoveredReceiver` all had a second explicit `objectWillChange.send()`. Removed those — each mutation now produces exactly one notification.

The bubble-up sink in `attachSession` (session → manager) is preserved; it is the only non-redundant explicit emit and is required so manager-bound views re-render when session state changes.

### 3. Defer the discovered-receiver picker mutation (`TBDisplaySenderContentView.swift`)

`.onChange(of: session.selectedReceiverID)` was calling `service.applyDiscoveredReceiver` synchronously from within the view-update phase, mutating `session.receiverIP` and `session.localTBIP` (both `@Published`) before SwiftUI had finished its current pass. Wrapping the mutation in `DispatchQueue.main.async { ... }` defers it to after the current view update completes and resolves the runtime warning.

## Test plan

- [ ] Build via Xcode (`TBDisplaySender` scheme) — verified locally, builds clean.
- [ ] Default behavior with the watchdog: stream for at least one full `replayd` interruption cycle (~30 min). Expect to see in `Console.app`:
      `TargetBridge: capture watchdog tripped — N.Ns since last frame, soft restart`
      `TargetBridge: watchdog (Ns without frames) — soft restart of capture pipeline`
      and the stream should recover automatically without the user noticing more than a brief blink.
- [ ] Confirm the receiver session remains connected across the watchdog-driven restart (`NWConnection` is preserved).
- [ ] At launch, observe `log show --predicate 'process == "TargetBridge"' --info --last 1m` — the `Publishing changes from within view updates` warnings should no longer appear.
- [ ] Toggle settings (Prevent display sleep, Auto-restart on wake, Verbose logging) repeatedly during streaming — UI remains responsive (no doubled notifications causing extra re-renders).
- [ ] Manually click the in-app "Restart capture" button — same soft-restart path as the watchdog, should behave identically.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -18,14 +18,27 @@ struct TBVirtualDisplayIdentity {
         usesDedicatedArrangementIdentity: false
     )
 
-    static func extendedDesktop() -> TBVirtualDisplayIdentity {
-        let random = UInt32.random(in: 0x0100...0xFFFE)
+    static func extendedDesktop(for profile: TBMonitorDisplayProfile) -> TBVirtualDisplayIdentity {
+        // Deterministic identity per receiver so macOS retains window placement
+        // and the saved extended-desktop arrangement across reconnects.
+        let key = "\(profile.receiverName)|\(profile.panelWidth)x\(profile.panelHeight)"
+        let hash = djb2(key)
+        let productLow = (hash & 0x00FF) | 0x01
+        let serialLow = (hash & 0xFFFE) | 0x0100
         return TBVirtualDisplayIdentity(
-            productID: 0x6000 | (random & 0x00FF),
-            serialNumber: 0x2027_0000 | random,
+            productID: 0x6000 | productLow,
+            serialNumber: 0x2027_0000 | UInt32(serialLow),
             displayNamePrefix: "TB Extend",
             usesDedicatedArrangementIdentity: true
         )
+    }
+
+    private static func djb2(_ input: String) -> UInt32 {
+        var hash: UInt32 = 5381
+        for byte in input.utf8 {
+            hash = hash &* 33 &+ UInt32(byte)
+        }
+        return hash
     }
 }
 
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderBuildInfo.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderBuildInfo.swift
@@ -1,5 +1,5 @@
 enum TBDisplaySenderBuildInfo {
     static let marketingVersion = "2.0"
-    static let buildNumber = "20260526205217"
+    static let buildNumber = "20260527204607"
     static let versionDisplay = "\(marketingVersion) + build \(buildNumber)"
 }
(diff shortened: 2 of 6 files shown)


swellweb:
Thanks a lot for this — the investigation here is very useful.

I’m pulling the safest/highest-value parts into the current release branch first:
- capture watchdog + soft restart
- deterministic extended-desktop virtual display identity
- deferring the discovered-receiver picker mutation out of the SwiftUI update phase

I’m leaving the broader `objectWillChange` cleanup for a second pass so it can be tested more carefully against the current release branch, which already has a lot of sender/UI changes on top.

Really appreciate the work here.

DrDavidL:
Still issues running overnight (although better!) and next set of mitigations in a new PR!

DrDavidL:
Superseded by #82, which contains this commit (`63ad77c`) plus the real-time PTS fix and the off-main-thread pipeline rework. Collapsing the stack into #82 so the whole long-session-stutter fix lands as a single PR.

swellweb:
Thanks again for this investigation and the fixes here. The safest/highest-value parts were adapted into the 3.0 release branch, and the release PR to main is now open. Really appreciate the work on this.

PR #98, opened 2026-06-05
fix(sender): key extended-display identity on receiver address

### LINKED ISSUE #92: BUG/REGRESSION: Can't create 2nd extended display
When I already have 1 VD (5K, extended) running:
- I add another session
- settings: extended, 5K
- click "connect"
- "Virtual display creation failed" (very quickly, after <100ms)

TargetBridge 3.0.0 ARM sender, Intel receiver

Interestingly, adding a 2nd display (mirror) works.

This is a regression, that worked in previous releases.

ThomasWaldmann:
Fixes #92

## Problem

Creating a **second extended display** fails immediately ("Virtual display creation failed", <100ms) when one extended VD is already running. Mirror is unaffected. This is a regression from 46b2c79.

## Root cause

The extended-desktop virtual display identity (`productID`/`serialNum`) is derived from `"<receiverName>|<panelW>x<panelH>"`:

- `receiverName` is the receiver's `SDL_GetDisplayName()`, and the panel size is hard-coded to `5120x2880` on the receiver.
- Two identical iMac panels (e.g. 5K 2017 + 5K 2020) therefore produce the **same key** → **same identity**.

macOS refuses to create a second `CGVirtualDisplay` with a duplicate vendor/product/serial, so the second extended display fails. Mirror works because it uses a separate fixed identity.

The commit that introduced this said the identity "matches the identity convention already used for the saved arrangement key" — but it didn't: the arrangement key discriminates by `receiverIP`, while the identity used `receiverName`.

## Fix

Derive the identity from the **same per-receiver discriminator the saved-arrangement key already uses** (the connection address, falling back to `receiverName`). Each receiver now maps to a distinct, stable identity across reconnects. Factored the discriminator into a shared helper so the identity and arrangement keys provably agree. Sender-only; no protocol/receiver change.

## Notes / caveats

- Verified it builds (`xcodebuild ... BUILD SUCCEEDED`).
- Edge case: if `receiverIP` is ever empty it falls back to `receiverName`, so two same-named receivers would still collide (unchanged from today). In a Thunderbolt-bridge setup the IP is always populated.
- Existing single-display users get a one-time change of the macOS per-display window-placement identity (was keyed on name, now on address).

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -18,11 +18,17 @@ struct TBVirtualDisplayIdentity {
         usesDedicatedArrangementIdentity: false
     )
 
-    static func extendedDesktop(for profile: TBMonitorDisplayProfile) -> TBVirtualDisplayIdentity {
+    static func extendedDesktop(receiverKey: String) -> TBVirtualDisplayIdentity {
         // Deterministic identity per receiver so macOS retains window placement
         // and the saved extended-desktop arrangement across reconnects.
-        let key = "\(profile.receiverName)|\(profile.panelWidth)x\(profile.panelHeight)"
-        let hash = djb2(key)
+        //
+        // `receiverKey` must uniquely identify the receiver (the caller derives it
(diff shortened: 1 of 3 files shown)


ThomasWaldmann:
Now 2 extended displays work, practically verified!

Review with care, mostly done by AI.

PR #99, opened 2026-06-05
Persist session count and per-session settings

### LINKED ISSUE #96: save settings for all sessions
The sender app should remember:
- the amount of desired sessions
- the settings (5K, extended, audio forwarding, brightness, ...) for these sessions

BTW, this is a very annoying missing feature. I usually want 5K/extended on 2 receivers: that are 7+ extra clicks for each trial of the app (and I try it a lot when working on it).

ThomasWaldmann:
## Summary

Fixes #96 — the sender now remembers both the **number of sessions** and **each session's settings** across launches.

Previously `init()` called `addSession()` exactly once, so every launch started with a single default-configured session and all per-session choices were lost.

## What's persisted

The sessions array is saved as JSON in `UserDefaults` (`fd.tbdisplaysender.sessions.v1`) and rebuilt on launch. Per session:

- Transport kind (Thunderbolt / Network Link)
- Local interface IP
- Discovered-receiver selection + receiver IP
- Capture source (mirror / extended)
- Stream profile (Standard … 5K)
- Audio forwarding
- Brightness
- Input gesture mode

The **session count** is implicit in the array, so 2 configured sessions restore as 2.

## Notes

- Writes are coalesced per main-loop tick and de-duplicated against the last encoded blob, so streaming's constant `objectWillChange` churn doesn't spam `UserDefaults`.
- Restored sessions are re-normalized against currently-enabled addons (`normalizeAddonState()`) and available local interfaces (`normalizeSessionInterfaces()`).
- **Input control role is intentionally excluded** — restoring it would make the app start grabbing/injecting keyboard & mouse on launch before any connection exists. It resets to `Off`.
- App-wide settings (large cursor, prevent display sleep, auto-restart on wake, verbose logging, language) were already persisted separately and are unchanged.

## Testing

`xcodebuild -scheme TBDisplaySender -configuration Debug` → **BUILD SUCCEEDED**.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -18,11 +18,17 @@ struct TBVirtualDisplayIdentity {
         usesDedicatedArrangementIdentity: false
     )
 
-    static func extendedDesktop(for profile: TBMonitorDisplayProfile) -> TBVirtualDisplayIdentity {
+    static func extendedDesktop(receiverKey: String) -> TBVirtualDisplayIdentity {
         // Deterministic identity per receiver so macOS retains window placement
         // and the saved extended-desktop arrangement across reconnects.
-        let key = "\(profile.receiverName)|\(profile.panelWidth)x\(profile.panelHeight)"
-        let hash = djb2(key)
+        //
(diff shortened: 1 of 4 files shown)


ThomasWaldmann:
Review with care, mostly AI generated.

Practically tested, it works great! So much less clicking now!

I rebase this on PR #98, because otherwise I would not be able to get 2 extended 5K displays due to the bug that was fixed there.

PR #102, opened 2026-06-05
fix(input): let relayed cursor cross onto adjacent displays

### LINKED ISSUE #97: Input-Dockstation: can't leave macbook screen with pointer
Macbook M3P: sender
iMac: receiver (iMac keyboard and touchpad shall control Macbook)

It mostly works, but when I use the iMac touchpad, I can not leave the Macbook screen to the right to enter the iMac screen with the pointer.

If I use the Macbook touchpad, this works as expected (extended display).

ThomasWaldmann:
## Problem

Fixes #97.

In Input-Dockstation **receiver-master** mode (e.g. iMac touchpad controlling a MacBook sender), the relayed pointer could not leave the sender's main screen to enter an adjacent virtual/extended display. Moving with the sender's *own* touchpad worked, since that path doesn't go through the relay.

## Root cause

Received mouse-move deltas are re-injected on the sender by `postLocalMouseMove`, which runs each target through `clampedMouseTarget`. That function clamped the target to the bounds of the **single display the cursor was currently on**, trapping the pointer on the sender's main display.

There was also a latent coordinate-system mismatch: the cursor location is in Quartz coordinates (top-left origin), but `screenFrame(containing:)` matched against `NSScreen.frame` (AppKit, bottom-left origin) — only correct by coincidence on the main display.

## Fix

- `clampedMouseTarget` now returns the target unchanged whenever it lands on **any** active display, allowing the pointer to cross onto adjacent screens — matching native multi-display behavior. It only clamps (to the current display) when the target would fall off every display, so the injected cursor can't get lost in a gap.
- Display lookups now use `CGDisplayBounds` (Quartz top-left origin), consistent with the cursor location.

## Testing

- `xcodebuild` of `TBDisplaySender` succeeds.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
@@ -1622,16 +1622,38 @@ final class TBDisplaySenderSession: NSObject, ObservableObject, Identifiable, @u
         CGEvent(source: nil)?.location
     }
 
+    // Bounds of every active display, in the Quartz global coordinate space
+    // (top-left origin) — matching CGEvent locations and CGWarpMouseCursorPosition.
+    // NSScreen.frame uses AppKit's bottom-left origin and must not be mixed in here.
+    private func activeDisplayBounds() -> [CGRect] {
+        var count: UInt32 = 0
+        guard CGGetActiveDisplayList(0, nil, &count) == .success, count > 0 else { return [] }
+        var ids = [CGDirectDisplayID](repeating: 0, count: Int(count))
+        guard CGGetActiveDisplayList(count, &ids, &count) == .success else { return [] }
+        return ids.prefix(Int(count)).map { CGDisplayBounds($0) }
+    }
+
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Please review carefully, mostly done by AI.

I tested it practically, it works!

swellweb:
Are you sure this it's valid also for receiver as primary(master) and reverse?

ThomasWaldmann:
i only tested the imac connected kbd and tp, can do more tests later.

ThomasWaldmann:
@swellweb I don't get the usecase of "This Mac is master". I thought the macbook kbd/tp could control the macOS running on the iMac - but how is that useful while I only see the screen output from the macbook on the iMac display and not the iMac's macOS desktop?

swellweb:
You can swith the view with mouse with swipe gesture and you can use the controller with the sender.

ThomasWaldmann:
I'm totally confused by that mode (in main branch), I can't test that.

So, can you test the code in this PR yourself, whether it broke something relating to that mode?

PR #107, opened 2026-06-08
Add sender-side CLI / automation (URL scheme + launch args)

kabirh:
Adds a focused, sender-side automation surface so TargetBridge can be driven without the GUI — for scripting, SSH, and connecting automatically on login or wake. (Following up on our discussion — thanks for the green light!)

  Two equivalent entry points, both handled in the existing app entry point and dispatched to `TBDisplaySenderService.shared` — they call the same `session.connect()` / `stop()` the GUI uses, so there's **no separate/parallel control
  path**:

  - **URL scheme:** `open "targetbridge://connect?receiver=auto&mode=mirror&preset=1440p"` / `open "targetbridge://disconnect"`
  - **Launch args:** `open -a TargetBridge --args --connect --receiver auto --preset 1440p` (login item / LaunchAgent)

  Options: `receiver` (auto | id | name | ip), `mode` (mirror | extended), `preset`, `transport`, `session`, `local-ip`. `receiver=auto` waits briefly for Bonjour and uses the first receiver; a raw IP/hostname bypasses discovery.

  Also included:
  - `cli/targetbridge` — a thin shell wrapper over the URL scheme.
  - `docs/Automation.md` — CLI, URL scheme, launch-at-login, SSH (`launchctl asuser … open`), and wake recipes.

  ### Scope
  Kept deliberately small per your guidance — practical sender-side connect automation that builds on the current model. Natural follow-ups (out of scope here): richer queryable control (status / wait-for-stream) and a long-lived control
  socket.

  ### Notes for review
  - Registers the `targetbridge` scheme via a small base `Info.plist` merged with Xcode's generated one. The three `INFOPLIST_KEY_*` usage strings move into that file because Xcode drops them once `INFOPLIST_FILE` is set (verified the
  built plist retains the generated keys + the scheme).
  - `project.pbxproj` is regenerated (xcodegen) to pick up the new file + the `INFOPLIST_FILE` setting.
  - Built and tested on Xcode 26.5 / macOS 14; verified end-to-end that both entry points drive the existing `connect()` and bring up the capture display.

  ### How to test
  Build + run the app, then:
  `open "targetbridge://connect?receiver=auto&mode=mirror&preset=1440p"` then `open "targetbridge://disconnect"`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ TargetBridge 3.0 turns the project into a more complete multi-Mac workspace:
 - Audio Relay: [docs/Features.md#audio-relay](docs/Features.md#audio-relay)
 - Input Dockstation, clipboard sync, and master/slave roles: [docs/Features.md#input-dockstation](docs/Features.md#input-dockstation)
 - Remote brightness control: [docs/Features.md#remote-brightness-control](docs/Features.md#remote-brightness-control)
+- CLI & automation (URL scheme, launch args, SSH, login/wake): [docs/Automation.md](docs/Automation.md)
 - Shared translations and language files: [docs/Features.md#shared-translations](docs/Features.md#shared-translations)
 - Thunderbolt networking extras (SSH/SFTP, file sharing, Internet Sharing): [docs/Features.md#thunderbolt-networking-extras](docs/Features.md#thunderbolt-networking-extras)
 
@@ -101,6 +102,7 @@ If you build from source, app outputs go into `build/` folder.
 ## Detailed Documentation
 
 - Feature overview: [docs/Features.md](docs/Features.md)
+- CLI & automation: [docs/Automation.md](docs/Automation.md)
 - Addon manifests and capability model: [docs/Addons.md](docs/Addons.md)
 - Audio transport internals: [docs/audio.md](docs/audio.md)
 - Hardware, cables, adapters, and Thunderbolt Bridge networking: [docs/Hardware.md](docs/Hardware.md)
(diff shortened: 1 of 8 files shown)


swellweb:
I pushed a small maintainer-side follow-up onto this PR branch to tighten two review points before merge:\n\n- invalid \ values now fail explicitly instead of silently falling back to session 1 or disconnecting everything\n- the automation connect path now refreshes local interfaces / discovery state before resolving \ and picking the local interface, which should make login/wake/SSH-driven flows more reliable\n\nI also rebuilt the Sender after the change to make sure the branch still compiles cleanly.\n\nThanks again — the overall direction here is still very good.

swellweb:
Small follow-up: I also pushed a maintainer-side fix on top of this PR branch to tighten two review points before merge:

- invalid `--session` values now fail explicitly instead of silently falling back to session 1 or disconnecting everything
- the automation connect path now refreshes local interfaces / discovery state before resolving `receiver=auto` and picking the local interface, which should make login/wake/SSH-driven flows more reliable

I rebuilt the Sender after the change as well, so the branch still compiles cleanly.

Thanks again — the overall direction here is still very good.

swellweb:
We’re currently testing this PR on our side as well.

I’m also making a few small maintainer-side adjustments to better align it with the current TargetBridge project structure and behavior, while keeping the overall scope the same.

So if you notice a couple of follow-up commits on this branch, that’s just us tightening the integration and validating it before merge.

kabirh:
Please feel free to edit, discard or integrate it as you see fit. I'm just happy to see this functionality make it in the main project and hope it expands the audience.

swellweb:
Thanks again for putting this together. I pulled the automation work into the current `feat/sender-cli-automation` release branch, tightened a couple of edge cases around session handling / discovery, updated the docs, and it is now included in `main` for the `v3.1.0` release flow.

Really appreciate the contribution here — this is a very useful addition to TargetBridge.

swellweb:
Closing this PR because the equivalent automation work has now been integrated directly into `main` for the `v3.1.0` release, with a few maintainer-side adjustments to keep behavior aligned and avoid regressions.

Thank you again for the contribution — it was genuinely helpful and it directly shaped this release.

PR #119, opened 2026-06-18
fix(receiver): only relay input while the shared display Space is active

tiehfood:
In receiverMaster, the global event tap forwarded every event regardless of which macOS Space was frontmost, so moving the mouse/keyboard on a receiver-only Space still drove the sender.

Gate forwarding on -[NSWindow isOnActiveSpace] of the receiver's content window (queried via NSApp, so it's purely spatial and independent of keyboard focus). When the shared window isn't on the active Space, events pass through locally and nothing is forwarded.

Test: in receiverMaster, switch to another Space on the receiver → sender cursor no longer moves; switch back → control resumes.


swellweb:

PR #120, opened 2026-06-18
fix(sender): persist the Input Dockstation role across launches

tiehfood:
The input master role was treated as runtime state and excluded from persistence, so it reset to off on every launch.

Persist it in PersistedSession (optional field, backward-compatible with older saved sessions) and restore it, enforcing the single-master invariant on restore.


swellweb:

PR #121, opened 2026-06-18
fix(sender): force dark color scheme on settings and about sheets

tiehfood:
These sheets use a fixed dark gradient background, but their text uses semantic colors (.primary/.secondary). In system Light mode those resolve to dark and render dark-on-dark / unreadable.

Pin both sheets to .preferredColorScheme(.dark) so the text stays legible regardless of the system appearance. No change in Dark mode.


swellweb:

PR #122, opened 2026-06-18
feat(sender): remember the virtual display resolution per receiver

tiehfood:
macOS doesn't reliably persist resolution for these hot-plugged virtual displays, and the session re-imposed the receiver-advertised default (e.g. HiDPI 2560×1440) on every connect.

Capture the user's chosen mode via a display-reconfiguration callback (keyed by virtual-display identity) and re-apply it on reconnect, matching pixel size so HiDPI vs Standard is preserved.

Test: set a non-default resolution while connected, reconnect → the choice sticks (one-time re-set per receiver, since nothing is stored on first connect).

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -97,16 +97,30 @@ final class ReceiverBackedVirtualDisplaySession {
             return false
         }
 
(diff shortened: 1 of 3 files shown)


swellweb:
Thanks for the contribution — really appreciated. I retargeted this to `3.2-dev` so new feature work can land there while `3.1-maint` stays focused on bug-fix releases.

swellweb:
Thanks for this useful contribution. It is now integrated into 3.2-dev, with the saved resolution scoped to the actual receiver (also in mirror sessions) and the display callback routed through the main actor to keep reconnect and teardown safe. The Sender test suite passes with these adjustments.

PR #123, opened 2026-06-18
perf(input): hand-roll input-event encoding to drop per-event JSONEncoder

tiehfood:
Mouse-move/drag events fire at display rate, and allocating a fresh JSONEncoder per event was the busiest allocator on the input relay path.

Emit the same JSON shape with a small hand-rolled encoder (mirrors the receiver's snprintf emitter). Wire format is unchanged, so it's fully compatible with existing receivers.


swellweb:
Thanks for working on this — really appreciated. I retargeted it to `3.1-maint` because we’re treating it as part of the current bug-fix line and want to keep those changes separate from the next feature release.

PR #124, opened 2026-06-18
feat(sender): add per-session configurable shortcut bindings

tiehfood:
Adds per-session bindings that map a trigger combo pressed on the master to an action combo injected on the slave (direction follows the input role). Edited via a click-to-record UI in the session settings and persisted per session.

The action is injected through in-process System Events (NSAppleScript), because macOS ignores synthetic CGEvents for protected symbolic hotkeys such as Space switching. Triggers must be combos macOS doesn't reserve (e.g. ⌃⌥←, not ⌃←); the action may be any combo.

Notes: UI strings are English-only (rest of app is localized); currently wires the receiverMaster direction. Requires Accessibility + a one-time "control System Events" Automation grant on the sender.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderContentView.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderContentView.swift
@@ -631,6 +631,12 @@ private struct TBDisplaySenderSessionSettingsSheet: View {
                             }
                         }
 
+                        if session.inputControlRole != .off {
+                            SurfaceSubcard {
+                                TBInputBindingsView(session: session)
+                            }
+                        }
+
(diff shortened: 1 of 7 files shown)


swellweb:
Thanks for the contribution — really appreciated. I retargeted this to `3.2-dev` so new feature work can land there while `3.1-maint` stays focused on bug-fix releases.

swellweb:
Thanks for the thoughtful shortcut-binding contribution. It is now integrated into 3.2-dev. I scoped it to the reliable Receiver is Master flow, added the Automation permission description and all UI translations, and changed modifier restoration so keys released during the System Events action can never become stuck. The Sender suite now passes with 68 tests, including coverage for binding matching.

PR #131, opened 2026-07-02
feat(receiver): add mock sender and loopback smoke test for hardware-free testing

jasontitus:
> **Stacked on #128** — includes its commits; review the last commit only. (The `hang` phase asserts #128's idle watchdog.)

## Motivation

Exercising the receiver previously required two Macs, a Thunderbolt cable, and an Apple Silicon sender — so failure paths (silent sender death, corrupt framing, mid-packet disconnects) were effectively untestable. This adds a protocol-level harness that needs none of that: everything runs against `127.0.0.1` on one machine, including CI-less local smoke coverage of the full decode pipeline.

## What's in it

- **`TBReceiverC/tests/mock_sender.py`** (stdlib-only Python 3) speaks the full wire protocol. Modes:
  - `handshake` — HELLO / heartbeat / TEARDOWN lifecycle
  - `stream` — PARAM_SETS + AVCC H.264 frames generated with the ffmpeg CLI (`testsrc2`)
  - `hang` — silent sender; the idle watchdog must reap it (~10s)
  - `badlen` — corrupt `0xFFFFFFFF` length prefix; parser must reject + disconnect
  - `drop` — abrupt mid-packet disconnect; receiver must return to waiting
- **`scripts/loopback_smoke.sh`** — builds the receiver, runs it windowed, drives all phases, asserts on the receiver's log, exits non-zero on failure.
- **`docs/Testing.md`** — documents every hardware-free test path (sender unit tests, receiver parser tests, mock sender, smoke script, same-Mac sender↔receiver) and how to stream the sender's connection log. Linked from the README.
- **`proto.h`**: corrects the stale `0x20` comment — the payload starts with a 1-byte codec marker (1=H.264, 2=HEVC) before the set count, as parsed by `build_extradata` in `decoder.c`.

## How tested

Full local run, all phases green: handshake acknowledged, corrupt length rejected + session dropped, silent sender reaped by the watchdog at ~10s, and **120 H.264 frames decoded and rendered** — the complete framing → decode → render pipeline with no Thunderbolt hardware.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,11 @@ jobs:
         run: |
           brew install ffmpeg sdl2 dylibbundler
 
+      - name: Run Receiver Parser Tests
+        run: |
+          cd TargetBridge-Receiver/TBReceiverC
+          make test
+
       - name: Build Receiver App
         run: |
           cd TargetBridge-Receiver
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ xcuserdata/
 # C build outputs
 *.o
 tbreceiver
+
+# C test outputs
+test_net_parser
+*.dSYM/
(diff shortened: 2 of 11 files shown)


swellweb:
Thanks for this valuable hardware-free test harness. The implementation is now integrated in 3.2-dev, including the mock sender, loopback smoke test, README link, and testing guide. I also added an assertion that decoded video reaches the renderer, so the smoke test catches a decoder that opens successfully but produces no frames. Parser tests pass (60 checks) and the full local loopback exercised handshake, corrupt framing, watchdog recovery, H.264 decode, and rendering.

PR #133, opened 2026-07-07
Allow receiver display sleep when idle

### LINKED ISSUE #132: Receiver prevents iMac display from sleeping even when idle
Environment
TargetBridge 3.1.0
Sender: MacBook Air M4 (macOS Sequoia)
Receiver: 2017 27" iMac 5K (macOS Ventura)
Problem

When the TargetBridge Receiver is running, the iMac display never turns off due to inactivity.

# Receiver prevents iMac display from sleeping when idle

## Environment

- TargetBridge 3.1.0
- Sender: MacBook Air M4 (macOS Sequoia)
- Receiver: 2017 27" iMac 5K (macOS Ventura)

## Problem

When the TargetBridge Receiver is running, the iMac display never turns off due to inactivity.

This occurs even when there is **no active TargetBridge connection**. Simply launching the Receiver is enough to permanently prevent display sleep until the application is quit.

## Expected behaviour

The Receiver should either:

- Allow normal display sleep when idle, or
- Only prevent display sleep while an active TargetBridge session is connected.

Ideally this would also be a user-configurable option.

## Actual behaviour

The display never sleeps while the Receiver process is running.

Running:

```bash
pmset -g assertions
```

shows:

```text
Assertion status system-wide:
   PreventUserIdleDisplaySleep    1

Listed by owning process:

pid 408(TargetBridgeReceiver):
PreventUserIdleDisplaySleep
named: "TargetBridge Receiver using SDL_DisableScreenSaver"
```

This suggests the Receiver is calling `SDL_DisableScreenSaver()` for its entire lifetime rather than only during an active TargetBridge session.

Quitting the Receiver immediately allows the display to sleep normally.

## Reproduction steps

1. Launch TargetBridge Receiver.
2. Do not connect a sender.
3. Wait longer than the configured display sleep timeout.
4. Observe that the iMac display never turns off.
5. Quit TargetBridge Receiver.
6. Observe that normal display sleep resumes.

## Notes

This appears independent of macOS display sleep settings and is reproducible every time.

freebo:
## Summary

Only prevent display sleep while an active TargetBridge session is connected.

Previously the Receiver called `SDL_DisableScreenSaver()` during application startup and only re-enabled it during application shutdown. As a result, the display would never sleep while the Receiver was running, even when idle.

This change:

- Removes `SDL_DisableScreenSaver()` from Receiver startup.
- Calls `SDL_DisableScreenSaver()` when a client connects.
- Calls `SDL_EnableScreenSaver()` when the client disconnects.
- Leaves the existing shutdown behaviour unchanged.

## Testing

Tested on a 2017 Intel 27" iMac running macOS Ventura 13.7.8.

Verified using:

```bash
pmset -g assertions
```

Results:

- Receiver running with no active connection → no `PreventUserIdleDisplaySleep` assertion.
- Sender connected → assertion present.
- Sender disconnected → assertion removed.
- Receiver quit → normal behaviour unchanged.

## Motivation

This fixes the issue where the Receiver prevented the display from sleeping whenever it was running, even if no Sender was connected.

Fixes #132

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/display.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/display.c
@@ -457,7 +457,6 @@ struct tb_display *tb_disp_create(int fullscreen) {
         fprintf(stderr, "[disp] SDL_Init: %s\n", SDL_GetError());
         return NULL;
     }
-    SDL_DisableScreenSaver();
 
     struct tb_display *d = (struct tb_display *)calloc(1, sizeof(*d));
     if (!d) return NULL;
--- a/TargetBridge-Receiver/TBReceiverC/src/main.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/main.c
@@ -1607,6 +1607,7 @@ static void close_client(struct app *a) {
     a->close_requested = 0;
     a->have_video_frame = 0;
     snprintf(a->input_control_mode, sizeof(a->input_control_mode), "off");
+    SDL_EnableScreenSaver();
     tb_receiver_refresh_input_capture(a);
     tb_disp_set_connection_state(a->disp, 0);
     tb_disp_set_cursor(a->disp, 0, 0, 1, 1, 0, 0);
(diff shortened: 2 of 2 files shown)


swellweb:
Thanks for this well-tested fix. We are bringing it into the 3.1 maintenance line and will resolve the small overlap with the receiver watchdog there, so both fixes ship together in 3.1.1.

PR #137, opened 2026-07-15
feat(display): add support for render matching to stream resolution

lionheo9:
## Summary

The receiver advertises a fixed 2560×1440 HiDPI profile, so the sender's virtual display always renders into a 5120×2880 backing store. Any capture preset below 5K therefore costs two lossy scale passes: ScreenCaptureKit downsamples 5120×2880 to the stream size, and the receiver upsamples the stream back to its panel.

This PR adds a per-session **Match render to stream** toggle (Output settings, Extended Desktop capture only) that sizes the virtual display's backing store to the active stream profile, so capture is 1:1 and only the panel-side scale remains. For example, with the Crisp preset the desktop runs at "looks like 1920×1080" HiDPI backed by 3840×2160 and is encoded with no rescale.

## Implementation notes

- `ReceiverBackedVirtualDisplaySession.create` accepts an optional `modeOverride`; macOS HiDPI backing stores are strictly 2×, so the mode is the stream preset halved in points. Falls back to the receiver profile if the requested backing store would exceed the advertised panel.
- Integrates with the per-receiver mode memory introduced on 3.2-dev: an explicit render-matching override takes precedence over a remembered manual mode (a stale pick would silently break the 1:1 capture the user asked for); with the toggle off, remembered-choice behavior is unchanged.
- The toggle is localized (en/it/de/zh) and disabled while streaming, matching the neighboring output settings.

## Testing

- `xcodebuild test` (TBDisplaySender scheme): 68/68 pass.
- Verified end-to-end with an Apple Silicon sender streaming to a 5K iMac receiver: with Crisp + the toggle on, the desktop reports 1920×1080 HiDPI, capture is 1:1, and text is visibly sharper than the double-scaled path.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
+++ b/TargetBridge-Sender/TBDisplaySender/ReceiverBackedVirtualDisplaySession.swift
@@ -5,6 +5,17 @@ extension CGVirtualDisplayDescriptor: @unchecked @retroactive Sendable {}
 extension CGVirtualDisplay: @unchecked @retroactive Sendable {}
 extension CGVirtualDisplaySettings: @unchecked @retroactive Sendable {}
 
+/// Pixel size of the mode handed to CGVirtualDisplay. With `settings.hiDPI = true`
+/// macOS synthesises a strictly 2x backing store, so a mode of (w, h) renders the
+/// desktop into a (2w, 2h) framebuffer and reports "looks like w x h" in Displays.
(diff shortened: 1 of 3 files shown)


swellweb:
Thank you for this thoughtful improvement and the clear write-up. I tested the integration on real Thunderbolt hardware: with Crisp, the virtual display correctly switches to the 3840x2160 backing store (1920x1080 HiDPI). It has now been merged into 3.2-dev. Really appreciate the contribution!

PR #138, opened 2026-07-15
feat(sender): streaming audio controls — volume key relay and local auto-mute

lionheo9:
## Summary

ScreenCaptureKit captures system audio upstream of the output device, so streamed audio always doubled onto the sender's local speakers, and once the sender was muted by hand the hardware volume keys only flashed the local mute bezel — receiver volume was adjustable solely from the app's slider. This PR makes streaming audio behave like a real output device:

- **Local auto-mute.** The default output mutes when a session starts streaming with audio and restores when the last one stops (claim-counted across sessions). Only a mute the app applied itself is undone — a pre-existing manual mute, or a deliberate unmute mid-stream, is respected.
- **Volume keys drive the receiver.** A consuming `NX_SYSDEFINED` event tap redirects sound-up/down/mute to the connected sessions' stream volume — 1/16 steps, 1/64 with shift+option, mute toggles and restores the previous level — which the receiver applies to its system volume with its own HUD. Events pass through untouched whenever nothing is streaming audio, so normal local volume behavior is unaffected.

## Implementation notes

- The tap needs Accessibility trust: it is created lazily, retried when the app becomes active (covers granting permission in System Settings and switching back), and a settings hint appears while trust is missing. Manually enabling the toggle triggers the system prompt.
- New settings toggle "Volume keys control the receiver while streaming" (default on, persisted, localized en/it/de/zh).
- Auto-mute uses CoreAudio (`kAudioDevicePropertyMute` on the default output); devices without a settable mute control are left alone.

## Testing

- `xcodebuild test` (TBDisplaySender scheme): 68/68 pass.
- Mute coordinator exercised against real hardware: unmuted → claim → muted → release → restored.
- Verified end-to-end streaming to a 5K iMac receiver: sender mutes on stream start and restores on stop; volume keys step the receiver's volume with its HUD while local keys behave normally when not streaming.

Generated with [Claude Code](https://claude.com/claude-code)

Cheeky claude adding the watermark.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderLocalization.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderLocalization.swift
@@ -327,6 +327,24 @@ enum TBDisplaySenderL10n {
         text("sender.toggle.show_menu_bar_icon", language)
     }
 
+    static func volumeKeysControlReceiver(_ language: TBDisplaySenderLanguage) -> String {
+        switch language {
+        case .italian: return "I tasti volume controllano il ricevitore durante lo streaming"
+        case .english: return "Volume keys control the receiver while streaming"
+        case .german: return "Lautstärketasten steuern den Empfänger beim Streamen"
+        case .chinese: return "串流时音量键控制接收端"
+        }
+    }
+
+    static func volumeKeysAccessibilityHint(_ language: TBDisplaySenderLanguage) -> String {
+        switch language {
+        case .italian: return "Richiede l'autorizzazione Accessibilità (Impostazioni di Sistema → Privacy e Sicurezza → Accessibilità)."
+        case .english: return "Requires Accessibility permission (System Settings → Privacy & Security → Accessibility)."
+        case .german: return "Erfordert die Berechtigung Bedienungshilfen (Systemeinstellungen → Datenschutz & Sicherheit → Bedienungshilfen)."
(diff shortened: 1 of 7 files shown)


aalpgiray:
I wonder if good solution to this problem is adding audio driver like microsoft teams or steam does so it looks like a speaker we select so we can switch between speakers or headphones like native os devices.

lionheo9:
At first I assumed that to be the default behavior. I opened sound output menu only to see “imac” unselected, and unable to be selected. 

Technically you can still use other sound outputs but it is a pain.

aalpgiray:
That I mac you see is continuity mode. And it doesn't work when you have multiple networks for some reason. And because of the TB cable we have multiple networks. Solving that issue could be one way to go perhaps but I was thinking creating our own device. Something like "TB - iMac".

aalpgiray:
image

I will create a pr soon and reference here

aalpgiray:
related: #151

swellweb:
Thank you for the Sender audio-controls work. The PR targets the retired 3.2 development branch, while 3.5.0 uses the current Audio Relay flow and has already been released with the tested connection and automation improvements.

I am closing this stale PR rather than carrying it forward without a full rebase and compatibility pass. A focused follow-up against the current development branch would be welcome when the audio-control scope can be tested end to end. Thank you again for the contribution.

PR #143, opened 2026-07-26
feat(sender): add brightness & volume sliders to the menu bar

aalpgiray:
image

## What

Adds **Brightness** and **Volume** sliders directly to the menu-bar status menu — one per connected session — so the receiver's real brightness/volume can be adjusted without opening the main window.

## Why

The receiver already applies real brightness (`DisplayServicesSetBrightness`) and volume, but today they're only reachable from the main window's session card. Surfacing them in the menu bar makes them a one-click adjustment — the same reliable, native-feeling pattern that apps like BetterDisplay / MonitorControl use (their own menu sliders), with none of the media-key / event-tap fragility.

## How

- Each slider is a custom-view `NSMenuItem` (SF Symbol icon + `NSSlider`) whose value is pushed live to `session.brightness` / `session.volume`, which already forward to the receiver. No event taps, no cursor/keyboard routing — it behaves identically regardless of external keyboards, Universal Control, etc.
- Sliders appear only for **connected** sessions; the volume slider only when audio is enabled.
- Also tidies the status menu: the verbose interface/IP line and per-session status lines move into a **Connection info** submenu, so the default menu is *status summary → sliders → actions*.
- Localized (en/it/de/zh).

Contained entirely in `TBDisplaySenderStatusItemController.swift`; no new files, no project changes.

## Testing

- `xcodebuild test` (TBDisplaySender scheme): **72/72 pass**.
- Verified end-to-end on a real receiver: dragging the menu-bar **Brightness** slider changes the receiver's real backlight; **Volume** changes the receiver's system volume.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderStatusItemController.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderStatusItemController.swift
@@ -7,6 +7,9 @@ final class TBDisplaySenderStatusItemController: NSObject {
     nonisolated(unsafe) private var statusItem: NSStatusItem?
     private var cancellables = Set<AnyCancellable>()
     private var hasActivated = false
+    // Retains the target objects for the current menu's sliders (NSSlider holds
+    // its target weakly). Cleared and rebuilt each time the menu opens.
+    private var sliderTargets: [TBMenuSliderTarget] = []
 
     init(service: TBDisplaySenderService) {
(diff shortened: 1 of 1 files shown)


aalpgiray:
After implementing audio driver for natural audio bridge 

file-703bab450402543cfed66648509efd17

I made this display only. 

file-22d12e925338e433afa6218f6e0d98d8

Will update this pr soon.

PR #146, opened 2026-07-27
fix(sender): stop receiver fullscreen flap when sender is idle

aalpgiray:
## Problem

With a sender app merely **open (not streaming)**, an idle receiver repeatedly toggles into its fullscreen "connecting" splash and back. It's distracting and makes the receiver look like it's receiving when nothing is being sent.

## Cause

`TBDisplaySenderManager` pushed the UI language to discovered receivers on **every** Bonjour discovery emission (`receiverDiscovery.$receivers.sink`). Bonjour re-emits the discovered set repeatedly (TXT refreshes, re-resolves), so the sender kept opening a throwaway TCP connection to each receiver just to send one `uiLanguage` packet. The receiver treats any accepted connection as a session start → fullscreen + "connecting" splash → the sender closes it → back to idle. Repeated pushes = the toggle.

## Fix

Track which receivers already received the current language (`languageSyncedReceiverIDs`) and push only to **newly-discovered** receivers; re-sync all only when the language actually changes. A stable discovery stream no longer reopens connections, so an idle receiver stays put.

Contained in `TBDisplaySenderManager.swift`; sender-only, no protocol change.

## Testing

- `xcodebuild test` (TBDisplaySender): **72/72 pass**.
- Behavioral: with the sender open and idle, the receiver no longer flaps fullscreen.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderManager.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderManager.swift
@@ -48,7 +48,9 @@ final class TBDisplaySenderService: ObservableObject {
         didSet {
             language.persist()
             sessions.forEach { $0.language = language }
-            pushLanguageUpdateToDiscoveredReceivers()
+            // Language actually changed: re-sync every discovered receiver.
+            languageSyncedReceiverIDs.removeAll()
+            syncLanguageToNewlyDiscoveredReceivers()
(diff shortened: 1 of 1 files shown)


aalpgiray:
Superseded by the receiver-side fix (idle-flap is fixed properly there): the receiver now ignores transient non-session connections instead of the sender trying to avoid opening them. Closing this partial sender-side attempt.

PR #147, opened 2026-07-27
fix(receiver): only show connecting splash for a real session (stop idle flap)

aalpgiray:
## Problem

With a sender app merely **open (not streaming)**, an idle receiver repeatedly flashes its fullscreen "connecting" splash and drops back — for ~2–3 s on sender launch, and more on multi-interface setups. Verified on a 2020 5K iMac.

## Cause

The receiver enters the fullscreen/"connecting" state on **any** accepted TCP connection. But the sender opens short-lived connections to discovered receivers during Bonjour discovery just to push the UI-language packet. On a multi-interface setup discovery churns while resolving each address, firing several of these transient connections — and the receiver treats each as a session start → fullscreen flash → close → repeat.

## Fix

Gate the fullscreen "connecting" state on a **real session** having started: set a `session_active` flag when a session packet arrives (`helloReceiver`, param sets, session ack, or a frame — all sent on a real connect, none by a language push), and stay on the windowed waiting screen otherwise. A transient language-only connection is now ignored; a genuine connect shows the splash exactly as before.

Receiver-only; no protocol change.

## Testing

- Receiver compiles clean (arm64 + x86_64).
- Verified on hardware (2020 5K iMac): sender open & idle → **no more flapping**; start casting → splash + video as before; 5K@60 raw passthrough still works.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/main.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/main.c
@@ -67,6 +67,10 @@ struct app {
     uint64_t last_recv_ms;      /* idle watchdog: last time the sender sent anything */
     int      close_requested;
     int      have_video_frame;
+    /* A real streaming session has begun (the sender sent a session packet, not
+     * just a transient probe like a UI-language push). Gates the fullscreen
+     * "connecting" splash so a bare/short-lived connection doesn't flash it. */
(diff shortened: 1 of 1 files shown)


aalpgiray:
I guess this is addressed.

PR #151, opened 2026-07-29
Use the receiver's speakers and microphone as macOS audio devices (optional addon)

aalpgiray:
file-33d874eb17ab038de97f89b0b45858aa

Adds an optional **Audio Driver** addon: the receiver Mac's speakers appear in the Sound menu like any other output — selectable system-wide or per app — and its microphone appears as an input. Audio previously reached the receiver only as the whole system mix, with no way to send one app there and keep the rest local.

**Optional and self-contained.** Off by default and marked experimental, because unlike Audio Relay it installs a system component and asks for admin. Nothing else depends on it, and it does not depend on Audio Relay — either addon alone gives a working audio path.

## Reviewing this

**95% of the diff is vendored [libASPL](https://github.com/gavv/libASPL)** (MIT, unmodified, `src/` and `include/` only), marked `linguist-vendored` so GitHub collapses it. Ours is ~900 lines across 20 files.

BlackHole solves a similar problem but is GPL-3.0 and this project is MIT. Implementing the AudioServerPlugIn interface directly instead of via a library would be roughly ten times the code.

**Why a HAL plug-in and not a DriverKit extension:** a plug-in loads ad-hoc signed, so this needs no Developer ID. A driver extension would need one plus an entitlement Apple grants on request.

## What installing the driver does

Nothing installs it automatically — it is a button in session settings, using the standard macOS authentication dialog rather than a separate installer package. That also lets the app *remove* it, and notice when the installed copy is stale.

| | |
|---|---|
| Adds | one bundle at `/Library/Audio/Plug-Ins/HAL/TargetBridge.driver` |
| Restarts | the system audio server — briefly interrupts audio in all apps |
| Ends | plug-in host processes holding **this** driver's binary (found via `lsof`, so other vendors' drivers are untouched) |
| Removal | `./uninstall.sh`, or the Remove button |

No launch agent, no daemon, nothing at login.

## Design points worth your attention

**The volume control reports without attenuating** — `ApplyProcessing` is a no-op. A stock loopback applies its slider to the samples before capture; combined with the forwarded hardware volume the level lands twice and dynamic range is lost. That is the entire reason for a custom control.

**Liveness is detected by probing, not listening.** The plug-in runs in a host process we do not control, where a leftover instance can hold a fixed port for minutes — `bind()` then fails with `EADDRINUSE` and never recovers. A connected UDP socket surfaces ICMP port-unreachable as `ECONNREFUSED` (RFC 1122 §4.1.3.3), so an unanswered probe *is* the "sender is gone" signal. Three strikes, then the device is **removed** rather than marked not-alive — macOS reliably moves the user elsewhere when a device disappears, the way real hardware does when unplugged; a not-alive device can be left selected and silent.

**The wire moves to Float32** — what CoreAudio mixes in and what real output devices advertise. Int16 forced a quantisation for nobody's benefit.

Since sender and receiver ship separately and are updated independently, this is negotiated in **both directions**, and both sides default to Int16 when the peer says nothing — a silent peer is an old peer:

| Direction | Announced in | Field | Absent means |
|---|---|---|---|
| Receiver → sender | display profile | `supportsFloat32Audio` | old receiver; sender sends Int16 |
| Sender → receiver | hello | `audioFormat` | old sender; receiver widens Int16 |

**Microphone latency targets 30 ms**, trimmed oldest-first past 60 ms. The two ends run on independent 48 kHz clocks with no shared reference, so drift accumulates one way and a merely-large buffer eventually runs full and stays full. This is the cheap tier of standard practice — PulseAudio and PipeWire resample toward a target latency, WebRTC's NetEq tracks a target delay and time-stretches — sharing the principle that matters: pick a target and correct toward it. Resampling would remove the occasional skip; worth adding only if measurement shows it matters.

**Every size derives** from 48 kHz, 2 channels and `sizeof(Float32)`, declared once per language in `TBAudioWireFormat.swift`, `Driver.cpp` and `proto.h`. Not tidiness — two hardcoded counts had already broken silently when the sample size doubled: a "0.5 second" ring became 0.25 s, and a "150 ms" backlog cap became 75 ms.

## Testing

Verified on a MacBook Pro M4 Pro sender and a 2020 Retina 5K iMac receiver:

- speakers and microphone both working, no drift over long sessions
- per-app routing via macOS Sound settings
- system volume slider and F11/F12 driving the iMac's hardware volume
- device withdrawn within ~3 s of force-quitting the sender, and on a clean stop
- install / update / remove from the app, including the stale-version prompt
- **system audio still correct with the addon disabled** — the pre-existing path
- quitting from the menu hands the output straight back, rather than waiting for the driver's probe to notice

**Known limitation:** apps that pick their output device explicitly (WhatsApp, Teams) ignore the system default and will not route here. Use their in-app picker.

Strings use the file's prevailing inline `switch language` pattern rather than the shared `TargetBridge-Shared/Languages/*.json` that `docs/Translations.md` describes — 105 such blocks across the sender against 8 files using the shared helper, so this follows local consistency. Happy to move them if the project is migrating. `check_localizations.sh` passes.

`docs/audio.md` is updated — it still described the Int16 pipeline and a byte count that had gone stale.


swellweb:
Thank you for the work on the optional audio-device addon. TargetBridge 3.5.0 retains the existing Audio Relay, but this HAL/device-driver approach needs a separate compatibility, security, packaging and long-term maintenance review before it can be merged safely.

I am closing this PR for now so it is not mistaken for a supported release feature. A future rebased proposal with a smaller scope and a dedicated test plan would be welcome. Thank you again for the contribution.

PR #152, opened 2026-07-29
Night Shift and True Tone toggles for the receiver's display

aalpgiray:
image

Adds Control-Center-style toggles to the menu bar for the receiver's **Night Shift** and **True Tone**, and drops the volume slider — with an audio device selected, macOS's own Sound slider already drives the receiver's hardware volume, so a second control here can only disagree with the system one.

Two-way: changing either on the receiver itself (Control Centre, System Settings) propagates back, so the toggles never show a stale state.

> **Builds on #143.** That PR's commit is included here, rebased onto current `main`. Merge #143 first, or merge this instead of it. The rebase also required filling in the French labels #143 predates — on current `main` its label switches are no longer exhaustive and it does not compile, which is worth knowing regardless of this PR.

## Private API, and how it's contained

Both features live in the private **CoreBrightness** framework, reached by `dlopen` plus the Objective-C runtime — the same approach this app already uses for DisplayServices brightness. A missing class or renamed selector degrades to "unsupported" rather than breaking the build or crashing on a future macOS. The receiver advertises `supportsNightShift` / `supportsTrueTone` in its display profile, and the toggles appear only when it does.

**`CBTrueToneClient` exposes both `activate`/`deactivate` and `setEnabled:`.** The first pair looks like the on/off switch and isn't — measured against the live framework, after `deactivate` the panel is unchanged and `enabled` still reports `1`. They activate the client *session*. `setEnabled:` is the real switch; `activate`/`deactivate` are kept only as a fallback.

**One deliberately fragile spot:** `CBBlueLightClient` has no `enabled` getter, only `getBlueLightStatus:`, which fills a private struct. It's read into an oversized zeroed buffer so a larger struct on a future macOS can't overflow, and a changed layout misreports the toggle rather than crashing. Worst case is a stale-looking toggle.

## Panel styling

The panel was ~240 pt against Control Center's ~340, and two symbols were simply the wrong picture.

Symbols were settled by **rendering every plausible SF Symbol and comparing against Control Center**, not by picking names that sounded right. Apple's family is half-circle → half-sun → full-sun for Dark Mode, Night Shift and True Tone — so Night Shift is a *sun with a dark half*, not a moon:

| | Before | After |
|---|---|---|
| Night Shift | `moon.fill` | `sun.lefthalf.filled` |
| True Tone | `sun.max` | `sun.max.fill` |
| Brightness ends | `sun.min` / `sun.max` | `.fill` variants |

True Tone's real glyph has lines through the disc and no SF Symbol does; Control Center appears to use a private asset, so `sun.max.fill` is the closest public match.

Width now comes from a single constant. Note the row width and the visible panel width are different numbers — a menu draws its own chrome around the item.

**The slider track stays grey, and that is a decision.** `trackFillColor` is set and menus ignore it. Overriding `NSSliderCell.drawBar` does produce the accent fill, but any drawing override opts the cell out of AppKit's modern slider rendering and the knob loses its pressed-state translucency. A control that *feels* wrong is worse than one that looks plain. The comment now says so, rather than leaving the next person to rediscover it.

## Accessibility

Fixed as part of this, since the custom views are ours:

- the slider carries its name — previously only its decorative glyphs had descriptions, so VoiceOver announced an unnamed slider flanked by two images
- the toggles use the checkbox role with a value that follows the state, instead of announcing a name with no on/off
- the captions are out of the accessibility tree rather than read as extra elements

## Conventions

Strings use the file's prevailing inline `switch language` pattern (105 such blocks across the sender, against 8 files using the shared `TBDisplaySenderL10n`). `docs/Translations.md` presents shared JSON as the mechanism, so this is a deliberate choice for local consistency — happy to move these keys into `TargetBridge-Shared/Languages/*.json` if the project is migrating that way. `check_localizations.sh` passes.

## Testing

Verified on a MacBook Pro M4 Pro sender and a 2020 Retina 5K iMac receiver: both toggles flip the iMac's panel, both reflect changes made on the iMac itself, and the toggles hide on a receiver that reports no support.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/Makefile
+++ b/TargetBridge-Receiver/TBReceiverC/Makefile
@@ -35,7 +35,7 @@ LDFLAGS  += -framework ApplicationServices -framework AppKit -framework CoreFoun
 endif
 
 C_SRC = src/main.c src/net.c src/decoder.c src/display.c src/tb_i18n.c
-OBJC_SRC = src/tb_gesture_bridge.m
+OBJC_SRC = src/tb_gesture_bridge.m src/tb_display_tweaks.m
 OBJ = $(C_SRC:.c=.o) $(OBJC_SRC:.m=.o)
 BIN = tbreceiver
 
--- a/TargetBridge-Receiver/TBReceiverC/src/main.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/main.c
@@ -18,6 +18,7 @@
 #include "display.h"
 #include "proto.h"
 #include "tb_gesture_bridge.h"
+#include "tb_display_tweaks.h"
 #include "tb_i18n.h"
 
 #include <SDL.h>
@@ -64,6 +65,13 @@ struct app {
     uint64_t last_fps_tick_ms;
     uint64_t last_fps_count;
     uint64_t last_ip_check_ms;
+    /* Last display-tweak state reported to the sender, so changes made on this
+     * Mac (Control Center, System Settings) propagate back and the sender's
+     * toggles stay truthful. -1 = nothing sent yet. */
+    int      reported_night_shift;
+    int      reported_true_tone;
+    uint64_t last_tweak_poll_ms;
+
     uint64_t last_recv_ms;      /* idle watchdog: last time the sender sent anything */
     int      close_requested;
     int      have_video_frame;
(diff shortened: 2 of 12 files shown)


swellweb:
Thank you @aalpgiray for the careful implementation, the hardware validation, and the attention to fallback behavior and accessibility.

I rebased the integration onto the current maintenance fixes, resolved only the generated build-number metadata, and verified the combined result: Sender tests (84/84), Receiver parser tests (60/60), and a full Receiver build all pass.

Merged into `dev-3.4.0` for the next feature release. Really appreciated.

Issue #64, opened 2026-05-24
Support Intel Macs as sender


ThomasWaldmann:
Maybe we can support recent Intel Macs as sender also.

swellweb:
Really, wow. It's fantastic news.

ThomasWaldmann:
It's based on #63.

If we get that working, nothing is holding us back from supporting Intel Macs as sender, right?

swellweb:
Exactly.

RiceCookPanda:
I'm looking forward to the recent Intel Macs sender being launched soon

fbnpn:
Would it maybe also be possible to enable Senders from Windows Clients (with built-in Thunderbolt of course)?

ThomasWaldmann:
@fbnpn Theoretically guess that it possible, but I think that should be in a different ticket, because it would need a lot of new windows-specific code, build automation, testing, etc.

swellweb:
I will create a project call TargetBridge Universal when I will back to work.
https://github.com/swellweb/targetBridge/discussions/125

preggocl:
Hi!!! I’ve been working on Intel Sender support and have now published the first public pre-release of my fork:

[https://github.com/preggocl/targetbridge-intel/releases/tag/v3.3.0-intel.1](https://github.com/preggocl/targetbridge-intel/releases/tag/v3.3.0-intel.1?utm_source=chatgpt.com)

It provides a native thin x86_64 Sender for Intel Macs running macOS 14 or later. I have validated it with:

Sender: 2020 Intel iMac 27-inch 5K, macOS Sequoia 15.7.7
Receiver: Intel iMac 21.5-inch 4K, macOS Monterey, using the unchanged upstream Receiver
Connection: direct Thunderbolt Bridge over bridge0
Confirmed Intel hardware encoders: H.264 GVA and HEVC AVE

The pre-release is intended to allow testing on additional Intel Mac models and macOS versions. So far, it has only been validated end-to-end on the configuration above.

In addition to adapting and packaging the Sender for Intel, I added a few features that may also be useful more generally:

Automatic Thunderbolt Bridge interface and Receiver-address inference using the actual bridge* address and netmask
Intel VideoToolbox hardware-codec selection and fallback diagnostics
4K HiDPI profiles prioritizing logical 2048×1152 and 2304×1296
Work 4K, Work 5K, Low Latency and Presentation quick profiles
Live telemetry with FPS history, codec, configured bitrate and resolution
[...]

PR #154, opened 2026-08-02
Build and publish Sender for x86_64

preggocl:
## Summary

- Build and test the Sender on both Intel (`x86_64`) and Apple Silicon (`arm64`) macOS runners.
- Verify the produced executable architecture in CI.
- Publish both Sender ZIP artifacts and provenance attestations in tagged releases.

## Validation

- `dev-3.4.0` was built natively as `x86_64` on an Intel iMac (macOS Sequoia 15.7.7).
- Sender unit suite passed: 84/84 tests.
- This PR intentionally contains no fork-specific bundle identity, UI, profile, localization, Receiver, or VideoToolbox behavior changes.

This is the first focused Intel Sender contribution discussed in #64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,14 +2,21 @@ name: CI Build Verification
 
 on:
   push:
-    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0 ]
+    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0, dev-3.4.0 ]
   pull_request:
-    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0 ]
+    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0, dev-3.4.0 ]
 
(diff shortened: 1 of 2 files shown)


swellweb:
Thank you for the focused PR and for keeping the first change limited to build and release coverage. I approved the fork workflow run and checked the results.

Both Sender builds reached `BUILD SUCCEEDED`, including the new Intel runner, so the matrix itself is working. The two Sender jobs then fail only in the architecture assertion: after `cd TargetBridge-Sender`, the packaging script writes the app to the repository-root `build/` directory. Could you update the CI check from:

```sh
file build/TargetBridge.app/Contents/MacOS/TargetBridge
```

to:

```sh
file ../build/TargetBridge.app/Contents/MacOS/TargetBridge
```

The release workflow is already correct because its artifact and zip steps run from the repository root. Once that small path fix is pushed, I will re-run the checks and proceed with the integration. Thanks again - this is exactly the kind of focused Intel contribution we hoped for.

PR #155, opened 2026-08-02
Build and publish Sender for x86_64

preggocl:
Hi! This replaces the earlier contribution after I renamed the fork branch for cleaner project naming.

## Summary

- Build and test the Sender on both Intel (`x86_64`) and Apple Silicon (`arm64`) macOS runners.
- Verify the produced executable architecture in CI.
- Publish both Sender ZIP artifacts and provenance attestations in tagged releases.
- Use the repository-root build path when checking the packaged Sender architecture.

## Validation

- `dev-3.4.0` was built natively as `x86_64` on an Intel iMac running macOS Sequoia 15.7.7.
- Sender unit suite passed: 84/84 tests.
- This remains limited to build and release support. It contains no fork-specific app identity, UI, display profile, localization, Receiver, or VideoToolbox behavior changes.

This is the focused Intel Sender build contribution discussed in #64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,14 +2,21 @@ name: CI Build Verification
 
 on:
   push:
-    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0 ]
+    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0, dev-3.4.0 ]
   pull_request:
-    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0 ]
+    branches: [ main, 3.1-maint, 3.2-dev, dev-3.3.0, dev-3.4.0 ]
 
 jobs:
   build-sender:
-    name: Build Sender (macOS 15 - Apple Silicon)
-    runs-on: macos-15
+    name: Build Sender (${{ matrix.os }} - ${{ matrix.arch }})
+    strategy:
+      matrix:
+        include:
+          - os: macos-15-intel
+            arch: x86_64
+          - os: macos-15
+            arch: arm64
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v6
@@ -26,6 +33,7 @@ jobs:
         run: |
           cd TargetBridge-Sender
           ./scripts/build_targetbridge_sender_app.sh
+          file build/TargetBridge.app/Contents/MacOS/TargetBridge | grep -q "${{ matrix.arch }}"
 
       - name: Run Sender Unit Tests
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         run: |
           cd TargetBridge-Sender
           ./scripts/build_targetbridge_sender_app.sh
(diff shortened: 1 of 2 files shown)


swellweb:
All checks are now green on both architectures. Thank you @preggocl for the quick correction and for keeping this contribution focused. Merging this as the first upstream Intel Sender build and release foundation.

GitSparTV:
Nice release! However, I tried to open it on MacBook Air I have in other place and it requires at least MacOS 14.
My MacBook Air 2017 no longer receiving updates and is stuck at 12.7.6.
I'm not aware what features are required from MacOS SDK, but is it possible to lower the requirement?

preggocl:
@GitSparTV  I’ll jump in here while this is being looked at upstream, since your compatibility issue with older macOS versions gave me a good reason to explore this a bit further.

I’ve now put together an experimental legacy compatibility branch of my Intel fork, aimed at Macs that are stuck on older macOS versions such as Monterey 12.7.x.

I was also able to test it on a 2017 Intel iMac 21.5-inch 4K running macOS Monterey 12.7.6, so it’s a pretty similar setup to yours. On that machine the legacy Sender launched correctly, created the virtual display and connected to the original TargetBridge Receiver.

One limitation for now is audio. The newer audio support depends on macOS features that simply aren’t available in Monterey, including newer ScreenCaptureKit audio capabilities and some of the Audio Relay-related pieces, so this legacy build currently focuses on getting the display side working properly first.

I’ve published a prerelease here if you’d like to give it a try:

https://github.com/preggocl/targetbridge-intel/releases/tag/v3.3.0-intel-legacy.1

It’s still experimental, so if you test it, any feedback would be really helpful, especially your exact Mac model, macOS version and what happens when you try to connect.
I’ll keep bringing over newer improvements where they can be made compatible with older macOS versions.

This branch is just meant to complement the main TargetBridge project and help keep some older Intel Macs useful a bit longer.

GitSparTV:
@preggocl Ah, bummer. My iMac, which is receiving, is at macOS 10.14.6 when the minimal supported is 11.
I specifically don't upgrade it, because of 32-bit support.
But I was able to run the sender app. It just... does nothing because i dont have working receiver :-(

preggocl:
Good news is you may not actually need to give up Mojave or your 32-bit apps 🙂
If your iMac’s system disk is APFS, you can add another APFS volume in the same container and install a newer macOS there, while keeping your existing Mojave installation untouched.
So you could keep booting into Mojave whenever you need your 32-bit software, and have something like Big Sur on a second volume just for TargetBridge Receiver. Big Sur is the minimum supported version for the Receiver.
APFS volumes share the free space in the same container, so you don’t normally need to permanently split the disk into fixed-size partitions either.
Then at startup you can simply choose whether to boot Mojave or the newer macOS.
That might be the easiest way to test the legacy Sender without losing the setup you’re intentionally keeping on Mojave.

GitSparTV:
@preggocl Thank you for reminding me about multi-macOS setup.
I've installed Monterey on iMac, connected with thunderbolt cable and nothing appears.
Sender sees receiver only using network. But even then it didn't event send a first frame. Sender blinked twice when creating display and then just stopped transmission. Receiver shows "waiting for the first frame".

Do I need to setup thunderbolt bridge/tunnel somehow?

PR #158, opened 2026-08-06
Experiment: getting a 5K Thunderbolt link to behave like a real monitor

aalpgiray:
## What this is

An experiment in driving a 2020 Retina 5K iMac as a **lossless** 5K60 second
display for an M4 Pro MacBook, over Thunderbolt.

Upstream streams HEVC, which is the right default: it is cheap and it works
anywhere. This branch asks a narrower question — if the link is a dedicated
~15 Gbps cable between two Macs you own, can you spend that bandwidth on
*fidelity* instead of compression, and still hold a locked 60?

**Draft.** It works, the numbers are below, and the branch is an integration
line rather than a reviewable change.

## Result

| | |
|---|---|
| capture cadence | **99–100%** of frames exactly one refresh apart |
| emission cadence | **94–98%** |
| receiver presentation | **90–100%**, best/worst 16.4 / 17.4 ms |
| frame rate | **60 fps locked** |
| dropped / incomplete / lost bands | **0 / 0 / 0** |
| `process` (capture callback → send) | **2.1 ms** of a 16.7 ms budget |
| pipeline latency | **~13–14 ms** end to end |

Lossless 4:4:4, 10-bit, no chroma subsampling, no generation loss.

## How it works

**Raw capture instead of transcoding.** ScreenCaptureKit hands over packed
BGRA (or ARGB2101010 for 10-bit) and it is never converted. No HEVC encode, no
4:2:0, no colour rounding.

**A codec built for this one link — TBD2.** Lossless tile-DPCM: 8×8 tiles,
per-tile bit widths, packed seeds, byte-aligned groups. Tiles are independent,
which is what makes everything downstream possible. Measured 6.5–7.8× on
desktop content, ~2.7× on photographic worst cases. A 58 MB frame becomes 7–8 MB.

**Both ends run on the GPU.** Encoding is three Metal passes on the sender;
decoding is a compute shader on the receiver's Radeon Pro 5300, which reaches
~2.3 ms per 5K frame against 166 ms on its CPU. The CPU on either machine never
touches a pixel.

**The receiver renders through Metal, not SDL.** SDL2 cannot present 10-bit;
a Metal plane with a BGR10A2 drawable can, plus colour management into Display
P3 and dithering into the 10-bit drawable.

**A true 10-bit virtual display.** `CGVirtualDisplay` is 8-bpc because it is
SDR — there is no depth property. Declaring HDR via the mode's
`transferFunction` gives a 16-bpc framebuffer, and the banding goes away.

**Frames are sliced into bands on the wire.** Because tiles are already
independent, a band is just a shorter frame written at a row offset — measured
byte-for-byte identical to a whole-frame encode at every slice count. Band *k*
decodes while band *k+1* is still arriving.

**Encoding never blocks capture.** The encoder submits and returns; the plan
step that used to run on the host between two GPU passes is now three kernels,
so a band is one command buffer with one completion and no host round trip.

That last one needed an identity that looks false at first. Each group's start
rounds up to a byte, which reads as strictly sequential — but every group base
is itself a multiple of 8, and for `a ≡ 0 (mod 8)`:

```
round8(a + b) == a + round8(b)
```

so the group bases are an ordinary prefix sum over *rounded* group totals.
Checked against the sequential algorithm on 20,000 random layouts before any
Metal was written.

**The receiver's speakers and mic become real audio devices** on the sender, via
a libASPL virtual driver that publishes and withdraws itself with the session.

**Both machines' logs land in one place.** The receiver's stderr is shipped to
the sender over the same link, because every measurement otherwise meant asking
someone to copy a terminal.

## What made it steady

Correctness came first; steadiness took longer, and most of it was configuration
and instrumentation rather than algorithms.

**ScreenCaptureKit was misconfigured.** `queueDepth` was 2 against Apple's
default of 8, and `minimumFrameInterval` was being used as a *target* when the
SDK header calls it a throttle: *"Set this to kCMTimeZero to capture at
display's native refresh rate."* A stalled SCK stream delivers **nothing, not an
idle frame**, so a queue stall and a compositor that drew nothing look identical
in the counters. Fixing both took capture from 94% to 100%.

**The receiver's reader must never stop reading.** A one-slot mailbox was
correct when a frame was one packet; slicing made it four, and a band cannot be
overwritten, so the reader waited for the renderer *inside the read loop* — on
the only thread draining the socket. A bounded queue replaced it.

**Two diagnostics were themselves the cost.** A 10-bit depth probe touching
~143k scattered locations across 59 MB was the entire `process` spike; the
telemetry formatting — two histograms and four `os_log` calls on the capture
thread, once per 240 frames — produced exactly one late frame per 240.

## Method

Measure at three points *together*: what the compositor drew, when packets
reached the socket, when pixels were presented. Any one alone points at the
wrong layer.

Measure a metric **at** the thing it names. `send(wall)` was sampled where the
frame was handed to the encoder; once encoding went async, packets left later
from a completion callback, so it read a clean 100% while the receiver visibly
bunched frames into pairs.

**Split an aggregate before theorising about it.** One `process` spike survived
three wrong theories; per-stage timers named it in a single run.

Every codec change is verified bit-exact against a reference C encoder at
1/2/4/8 bands and both depths, and every band decodes back to its source pixels.

## What did not work

Kept because each closes off a direction that otherwise looks obvious.

**Oversampling is impossible on a virtual display.** Run it at 120 Hz and send
only 60 and you should get fresher frames free. You do not: the display holds
the 120 Hz mode, but the compositor only *produces* 120 when you *consume* 120.
With no hardware scanout forcing a beat, refresh is demand-driven.

**Damage rects cannot work as a bounding box.** The WindowServer's damage is
spatially scattered — cursor, clock, a window — so one rectangle covers nearly
the whole screen. Measured 4 rect-frames in 240, averaging 85% of the frame.
Multi-region encoding is not an optimisation of that, it *is* the feature.
Implemented end to end and left off by default.

**The pixel-buffer lock has to stay.** Removing it is correct on paper and made
the sender's own numbers perfect — while receiver presentation fell from 100% to
80%, because the lock was accidental backpressure spacing submissions.

**8 slices has failed twice**, for two different causes. Worth a few ms of
overlap and nothing visible.

## Not done

Multi-region damage encoding; deliberate submission pacing to replace the lock's
accidental version; a rare `process` spike with none of our code on the stack.


aalpgiray:
I have to inform everyone reading this. I am only a web developer so these experiments are build completely with ai help.

swellweb:
Thank you for exploring the 5K Thunderbolt monitor experiment. For 3.5.0 we intentionally kept the release limited to the smaller, production-tested connection and automation improvements, so we are not merging this broader experiment as-is.

I am closing it to avoid leaving a stale integration proposal open. If you would like to continue, a smaller rebased PR with a clearly bounded test plan would be very welcome. Thanks again for the investigation.

PR #157, opened 2026-08-06
[RFC] Automatic iMac monitor mode — integrated prototype

Betafer:
## Integration preview — not ready for line-by-line review

I discovered TargetBridge while looking for a practical way to reuse a 21.5-inch 2017 Retina 4K iMac as the display for a Mac mini M4. Apple's Target Display Mode is not available on this iMac, while ordinary screen sharing was too slow for comfortable daily use.

I really liked TargetBridge's approach: instead of pretending the iMac is a physical DisplayPort monitor, it creates a virtual display on the Sender and carries a low-latency video stream to the Receiver. Starting from that design, I tested it with Thunderbolt, USB4/USB-C networking, direct Ethernet and Wi-Fi, then gradually added the pieces that made the setup feel closer to a real monitor: automatic startup, reliable Stop/Start, cable selection, 4K60, optional input control and separate audio controls.

This draft PR is the fully integrated and hardware-tested prototype. It is intentionally kept as a reference, not as a request to merge all 62 changed files at once. I plan to submit the useful parts as smaller, focused PRs so the project author can choose what fits TargetBridge.

## Proposed functional split

| Proposal | Utility | Independence |
| --- | --- | --- |
| [#160 — Native Retina 4K + stable 60 FPS](https://github.com/swellweb/targetBridge/pull/160) | Adds the 4096 × 2304 profile, receiver-native capture and the paced HEVC pipeline | Extracted as a standalone draft PR |
| [#159 — Thunderbolt, USB4, Ethernet and Wi-Fi path selection](https://github.com/swellweb/targetBridge/pull/159) | Identifies each macOS network path, measures viable links and selects the fastest one | Extracted as a standalone draft PR |
| [#161 — Reliable headless monitor lifecycle](https://github.com/swellweb/targetBridge/pull/161) | Login-safe startup, persistent user Stop, reconnect and headless display wake | Extracted as a standalone draft PR |
| Receiver-side Start/Stop and path controls | Lets the iMac operate a headless Sender without a second Mac mini display | Optional monitor-mode layer |
| iMac keyboard, mouse and clipboard control | Optional input relay with permission, focus and lock-up safeguards | Optional |
| Audio and display controls | Independent Sender/Receiver volume plus Receiver display-state reporting | Optional |
| [#162 — Receiver notification shielding](https://github.com/swellweb/targetBridge/pull/162) | Keeps local iMac banners behind the active fullscreen video and restores normal level on Stop | Extracted as a standalone draft PR |
[...]

PR #159, opened 2026-08-07
Add measured USB4 and network path selection

Betafer:
## Context — a short story

I discovered TargetBridge while looking for a practical way to reuse a 21.5-inch 2017 Retina 4K iMac as the display for a Mac mini M4. Apple's Target Display Mode is unavailable on this iMac, while ordinary screen sharing was too slow for comfortable daily use.

I really liked TargetBridge's design: it creates a virtual display on the Sender and carries a low-latency encoded stream to the Receiver, instead of pretending that the iMac is a physical DisplayPort monitor. While testing it with Thunderbolt, USB4/USB-C, direct Ethernet and Wi-Fi, I found that the hardest part for a non-technical user was not the stream itself, but making sure it used the best cable instead of silently falling back to Wi-Fi.

This PR isolates that one improvement from the larger integrated prototype in #157, so it can be reviewed and accepted independently.

## What this PR adds

- Discovers separate Receiver endpoints for:
  - Thunderbolt Bridge
  - direct USB/USB4 networking
  - Ethernet
  - Wi-Fi
- Identifies the matching local interface without assuming that Wi-Fi is always `en0` or `en1`.
- Performs a bounded, protocol-valid 16 MiB transfer over each candidate path.
- Binds every probe to the exact local interface and measures application-level throughput plus connection latency.
- Selects the measured fastest working path; interface names are used only as a tie-breaker.
- Adds explicit preferences: `auto`, `wired`, `thunderbolt`, `usb`, `ethernet`, and `wifi`.
- Exposes the selection through `targetbridge connect --path ...` and the existing URL/launch-argument automation.
- Keeps legacy behavior unchanged when `--path` is omitted.

## USB4 clarification

This does **not** turn the iMac into a native USB-C/DisplayPort sink and does not claim a raw 40 Gbit/s video path.

When macOS creates a direct USB-C/USB4 network interface between the two Macs, it normally appears as USB-NCM with a self-assigned `169.254.x.x` address. TargetBridge continues to use its encoded network protocol over that interface. The selector measures the throughput TargetBridge can actually obtain, rather than trusting the cable's advertised link rate.

The same mechanism deliberately supports direct Ethernet and Thunderbolt Bridge, because the fastest real path depends on both Macs, adapters, macOS routing and the cable.

## Receiver behavior

The Receiver advertises its Thunderbolt, USB/link-local, Ethernet and Wi-Fi IPv4 addresses as separate Bonjour TXT fields. Probe packets use the existing `TEST_DATA` packet type, which the Receiver discards without marking a video session active, so path selection does not start capture or flash the fullscreen session UI.

## Examples

```bash
targetbridge connect --receiver auto --path auto
targetbridge connect --receiver auto --path wired
targetbridge connect --receiver auto --path usb
```

## Intentionally not included

To keep this review focused, this PR does not include:

- automatic retry or headless startup
- Receiver-side Start/Stop controls
- Retina 4K/60 capture changes
- keyboard, mouse or clipboard forwarding
- audio/display controls
- installer or packaging changes

Those parts remain described in #157 and can be proposed as separate PRs by utility.

## Validation

Tested on a Mac mini M4 and a 21.5-inch 2017 Retina 4K iMac with direct USB4/USB-C, Thunderbolt Bridge, Ethernet and Wi-Fi paths.

- Sender: 96 XCTest tests passed
- Receiver networking: 67 C checks passed
- Receiver: full build completed successfully
- CLI: shell syntax and help output verified
- `git diff --check`: clean

No credentials, privileged helper, kernel extension or third-party dependency is added.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/Makefile
+++ b/TargetBridge-Receiver/TBReceiverC/Makefile
@@ -31,7 +31,7 @@ LDFLAGS  += $(shell $(PKG) --libs libavcodec libavutil sdl2)
 
 UNAME_S  := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
-LDFLAGS  += -framework ApplicationServices -framework AppKit -framework CoreFoundation -framework CoreGraphics -framework CoreText -framework VideoToolbox -framework CoreMedia -framework CoreVideo -framework IOSurface -framework CoreServices -framework CoreAudio
+LDFLAGS  += -framework ApplicationServices -framework AppKit -framework CoreFoundation -framework CoreGraphics -framework CoreText -framework VideoToolbox -framework CoreMedia -framework CoreVideo -framework IOSurface -framework CoreServices -framework CoreAudio -framework SystemConfiguration
 endif
 
 C_SRC = src/main.c src/net.c src/decoder.c src/display.c src/tb_i18n.c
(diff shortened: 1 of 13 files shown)


swellweb:
Thank you for this focused contribution. The measured connection-path selection has been merged and released in TargetBridge 3.5.0 after local, live Thunderbolt and Apple Silicon/Intel CI verification.

PR #161, opened 2026-08-07
Make monitor sessions headless and keep user Stop persistent

Betafer:
## A short story

Thank you for TargetBridge — I really enjoyed its architecture and the possibility it gives old iMacs.

My test setup is a Mac mini M4 with no permanent display and a 21.5-inch 2017 Retina 4K iMac as the Receiver. Once the video was working well, the remaining practical problem was startup: after a reboot or a temporary cable failure, the Mac mini still needed another monitor; conversely, an automatic retry loop could make the iMac impossible to stop. This change is the lifecycle part of that experiment.

It is split from the integrated prototype in #157 and is independent from both USB4 path selection (#159) and Retina 4K capture (#160), so it can be reviewed on its own.

## What this PR adds

- launch-argument automation is handled during app initialization, without depending on SwiftUI opening or restoring the main window;
- `--retry` monitors an active session and retries transient startup, Receiver and cable failures;
- GUI Stop, automation disconnect, Command-Q and menu-bar Quit cancel retry and remove a small `enabled` marker;
- internal recovery uses a separate stop reason and does not clear the user's monitor-mode choice;
- the final teardown reason is delivered before the connection closes, so a Receiver can distinguish a user Stop from a broken link;
- Quit waits for teardown delivery, with a two-second safety ceiling;
- capture wakes and holds the headless graphical session before virtual-display enumeration;
- Screen Recording permission failures and first-frame failures suspend retry instead of looping forever;
- ScreenCaptureKit is polled briefly after wake, then an online display can use the existing direct capture fallback;
- menu actions run only after AppKit confirms the status menu has closed, preventing an invisible menu layer from swallowing clicks.

## State marker

The optional marker is:

`~/Library/Application Support/TargetBridge/Sender/enabled`

This PR only defines its lifecycle semantics. A later, separate installer PR can use it as a LaunchAgent `PathState`: presence means automatic monitor mode is wanted; removal means the user explicitly stopped it. No LaunchAgent or packaging files are included here.

## Validation

- Xcode test suite on the standalone branch: **90 passed, 0 failed, 0 skipped**;
- six focused tests cover marker paths, user/permission/capture stop behavior and retry-flag parsing;
- CLI syntax and `git diff --check` passed.

The integrated hardware prototype was also exercised across three Sender restarts and one Receiver restart, recovering in approximately 3.1–7.4 seconds. That run is supporting hardware evidence; the reproducible validation for this intentionally standalone PR is the suite above.

## Deliberately out of scope

- selecting or measuring USB4/Thunderbolt/Ethernet/Wi-Fi paths;
- the Retina 4K profile and frame-rate changes;
- Receiver-side Start/Stop UI and remote recovery transport;
- keyboard, mouse or clipboard relay;
- installer and LaunchAgent creation.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderApp.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderApp.swift
@@ -1,3 +1,4 @@
+import AppKit
 import SwiftUI
 
 @main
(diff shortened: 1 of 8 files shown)


swellweb:
Thank you for this focused contribution. The resilient headless lifecycle and persistent user Stop behavior have been merged and released in TargetBridge 3.5.0 after local, live Thunderbolt and Apple Silicon/Intel CI verification.

PR #165, opened 2026-08-10
Receiver: preserve key and button releases under queue pressure

Betafer:
## Context

I have been using TargetBridge to reuse a 2017 4K iMac as the display and input station for a Mac mini M4 over a direct USB4/Thunderbolt Bridge link. It has been a great fit for that setup. While testing Receiver Master input and repeated reconnects, I found a small failure mode worth isolating from the larger feature work.

## Problem

The SDL fallback input queue is a fixed 128-event ring. When it fills, it currently overwrites the oldest event without considering its type. Under a burst of mouse motion or scrolling, this can discard a key-up or mouse-button-up event and leave the Sender with a logically held key or button.

## Change

- Move the queue policy into a small standalone module.
- Treat motion, drag and scroll events as lossy.
- Under pressure, evict the oldest lossy event before any key/button lifecycle event.
- If the queue contains only critical events, collapse to the existing deactivate-control fail-safe so the Sender releases all held input.
- Keep the queue bounded and preserve FIFO order in normal operation.

There is no protocol change.

## Validation

- Receiver parser tests: 67 checks passed.
- Input queue tests: 562 checks passed, including wraparound and key/button release survival.
- Queue tests also passed with AddressSanitizer and UndefinedBehaviorSanitizer.
- Full Receiver builds passed for arm64 and x86_64.
- Hardware-tested on a 2017 21.5-inch 4K iMac and Mac mini M4 over a direct USB4/Thunderbolt Bridge link.
- Three Receiver restart cycles and two Sender restart cycles reconnected successfully; input-state probes reported no held modifiers or mouse buttons after every cycle.

I am opening this as a draft so the queue policy and fail-safe behavior can be reviewed independently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,5 @@ tbreceiver
 
 # C test outputs
 test_net_parser
+test_input_queue
 *.dSYM/
--- a/TargetBridge-Receiver/TBReceiverC/Makefile
+++ b/TargetBridge-Receiver/TBReceiverC/Makefile
@@ -34,7 +34,7 @@ ifeq ($(UNAME_S),Darwin)
 LDFLAGS  += -framework ApplicationServices -framework AppKit -framework CoreFoundation -framework CoreGraphics -framework CoreText -framework VideoToolbox -framework CoreMedia -framework CoreVideo -framework IOSurface -framework CoreServices -framework CoreAudio -framework SystemConfiguration
 endif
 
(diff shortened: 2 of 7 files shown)


swellweb:
Thank you for isolating this input-queue edge case and for the thorough coverage. It is now merged into maint-3.5 for the next 3.5.1 maintenance release.

PR #166, opened 2026-08-10
Sender: use low-latency cursor overlay in Receiver Master

Betafer:
## Context

While using Receiver Master on a 2017 4K iMac connected to a Mac mini M4, I found that cursor handling is better kept on TargetBridge's existing low-latency cursor channel instead of being baked into the video stream. This is a small, focused extraction from the hardware-tested setup.

## Problem

With Receiver Master enabled and Large Cursor disabled, the Sender still includes the macOS cursor in captured frames. A Receiver that presents its control cursor can therefore show both the delayed video cursor and the local/overlay cursor.

## Change

- Treat Receiver Master as a low-latency cursor-overlay mode.
- Exclude the captured cursor in both ScreenCaptureKit and direct CGDisplayStream paths.
- Start the existing cursor packet updates for Receiver Master even when Large Cursor is off.
- Soft-restart capture only when a live role change crosses between captured-cursor and overlay-cursor modes.
- Keep Off and Sender Master behavior unchanged; Large Cursor continues to force the overlay as before.

There is no protocol change.

## Validation

- All 110 Sender unit tests passed.
- Added focused coverage for role-to-cursor-mode selection and live role transitions.
- Hardware-tested on a 2017 21.5-inch 4K iMac and Mac mini M4 over a direct USB4/Thunderbolt Bridge link.
- Manual Receiver Master testing confirmed a single responsive cursor.
- Repeated Sender and Receiver reconnect cycles returned to an established session without stale input state.

I am opening this separately as a draft so cursor transport can be reviewed without the other Receiver/input changes.

### DIFF
--- a/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
+++ b/TargetBridge-Sender/TBDisplaySender/TBDisplaySenderService.swift
@@ -350,6 +350,18 @@ enum TBInputControlRole: String, CaseIterable, Identifiable {
     case receiverMaster
 
     var id: String { rawValue }
+
+    func usesLowLatencyCursorOverlay(largeCursorEnabled: Bool) -> Bool {
+        largeCursorEnabled || self == .receiverMaster
+    }
+
+    func changesCursorCaptureMode(
+        from previousRole: TBInputControlRole,
+        largeCursorEnabled: Bool
+    ) -> Bool {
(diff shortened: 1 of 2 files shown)


swellweb:
Thank you for the focused Receiver Master cursor fix and the accompanying tests. It is now merged into maint-3.5 for the next 3.5.1 maintenance release.

PR #169, opened 2026-08-15
fix(receiver): advertise the physical display profile

preggocl:
## Summary

The Receiver currently advertises a fixed 5120×2880 / 2560×1440 HiDPI profile regardless of its actual panel. This makes non-5K receivers appear as 5K targets and can force inappropriate virtual-display geometry.

This change derives the profile from the screen hosting the Receiver window:

- obtains native panel pixels from the matching macOS `NSScreen`;
- keeps SDL's logical desktop mode separately;
- reports HiDPI only when the native backing dimensions are at least 2× the logical mode;
- sends native geometry for non-Retina displays, including 1920×1080.

The existing drawable fallback remains in place if native screen lookup is unavailable.

## Verification

- `make test` passes: 67 packet-parser checks and 13 receiver-profile checks.
- Added profile coverage for 5K Retina, 4K Retina, non-Retina 1080p, and invalid dimensions.
- Objective-C bridge compiles with `-mmacosx-version-min=11.0` in syntax-only validation.

A full local Receiver link was not possible in this checkout because the required FFmpeg/SDL pkg-config packages are not installed on this host.


swellweb:
Thank you for the careful physical-display profile fix and its focused test coverage. It is now merged into maint-3.5 for the next 3.5.1 maintenance release.

PR #170, opened 2026-08-15
fix(receiver): fall back to software video decode

preggocl:
## Summary

The Receiver already asks FFmpeg for a VideoToolbox decode device, but it drops frames when the selected decoder produces a software pixel format. This can happen when VideoToolbox is unavailable or rejects an individual stream on older hardware.

This change keeps the VideoToolbox fast path intact and adds a safe fallback:

- retains the hardware format per decoder instance instead of a process-global value;
- accepts a software format when FFmpeg retries format selection;
- converts software-decoded frames to NV12 with `libswscale` before handing them to the existing renderer;
- logs the fallback only when it is selected.

## Verification

- `make test` passes: 67 packet-parser checks.
- The fallback path was exercised in an Intel legacy Receiver build receiving H.264 streams after VideoToolbox was unavailable.
- Full local linking of this upstream worktree was not possible because its FFmpeg/SDL development packages are not installed on this host; the Makefile now explicitly includes `libswscale` for the added conversion dependency.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/Makefile
+++ b/TargetBridge-Receiver/TBReceiverC/Makefile
@@ -23,11 +23,11 @@ MACOSX_DEPLOYMENT_TARGET ?= 11.0
 export MACOSX_DEPLOYMENT_TARGET
 
 CFLAGS   += -O2 -g -Wall -Wextra -Wno-unused-parameter -std=c11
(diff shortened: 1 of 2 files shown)


swellweb:
Thank you for adding this safe software-decode fallback for older hardware. It is now merged into maint-3.5 for the next 3.5.1 maintenance release.

PR #172, opened 2026-08-21
fix(receiver): release audio device while idle

Betafer:
## Summary

Keep the Receiver audio output device closed while it is waiting for a Sender, and open it lazily on the first audio packet.

On macOS, an open SDL output device can cause coreaudiod to hold PreventUserIdleSystemSleep even when the callback only emits silence. It also touches the system output unnecessarily for sessions where audio relay is disabled.

## Behavior

- no audio device is opened while the Receiver is idle
- the device opens on the first audio frame and playback starts after that frame is queued
- the device and queued samples are released when the client disconnects
- failed device opens are rate-limited to one retry every five seconds
- existing audio packet and buffering behavior is otherwise unchanged

## Verification

- Receiver make test: 67 parser checks, 562 input-queue checks, 13 receiver-profile checks; all pass
- verified on a 2017 Intel iMac: audio opens during an active relayed-audio session and the Receiver no longer owns a coreaudiod sleep assertion after disconnect

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/main.c
+++ b/TargetBridge-Receiver/TBReceiverC/src/main.c
@@ -108,6 +108,7 @@ struct app {
     int      input_tap_consumes_events;
 
     SDL_AudioDeviceID audio_device;
+    uint64_t audio_retry_after_ms;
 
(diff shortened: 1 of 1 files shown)


swellweb:

PR #173, opened 2026-08-21
fix(receiver): guard idle watchdog clock underflow

### LINKED ISSUE #156: Receiver idle watchdog underflow: "no data from sender for 18446744073709551564 ms" while decoding 30 fps
## Summary

The Receiver incorrectly disconnects an active session while it is still continuously receiving video.

The disconnect is initiated by the Receiver.

This appears to be caused by an unsigned elapsed-time underflow inside the stale-session watchdog.

The Receiver reports

```
no data from sender for 18446744073709551564 ms
```

while it is still decoding approximately 30 fps.

This causes a false stale timeout and closes the TCP connection.

---

## Environment

TargetBridge

- Sender 3.4.2
- Receiver 3.4.2

Sender

- Mac Studio
- macOS Tahoe 26.5.2

Receiver

- iMac Pro 2017
- macOS Sequoia 15.7.7

Transport

- Thunderbolt Bridge
- Direct Thunderbolt 4 cable

---

## Receiver log

Immediately before disconnect

```
[main] 29 fps
[main] 30 fps
[main] 29 fps
[main] no data from sender for 18446744073709551564 ms; closing stale session
[main] client disconnected
```

The Receiver is clearly still decoding video immediately before disconnect.

Therefore the Sender is still transmitting.

This is not a genuine idle timeout.

See attached Receiver log.

---

## Why this strongly suggests an unsigned underflow

The elapsed value

```
18446744073709551564
```

equals

```
UINT64_MAX - 51
```

which corresponds to

```
-52 ms
```

interpreted as an unsigned 64-bit integer.

This strongly suggests something equivalent to

```cpp
uint64_t elapsed = now_ms - last_data_ms;
```

where

```
now_ms < last_data_ms
```

for a short period.

Instead of

```
elapsed = -52 ms
```

the calculation becomes

```
elapsed = 18446744073709551564 ms
```

which immediately satisfies

```
elapsed > TB_SENDER_IDLE_TIMEOUT_MS
```

and closes the session.

---

## Previous observations

This has now happened repeatedly.

Every disconnect follows the same pattern.

Receiver

```
tcp_usrclosed
```

↓

Sender

```
Connection reset by peer
```

↓

Virtual display removed

Therefore the Receiver appears to initiate every disconnect.

---

## Where I believe the bug is

Looking at the Receiver source,

```
TBReceiverC/src/main.c
```

contains the idle watchdog responsible for stale-session detection.

The watchdog appears to compare

```
now_ms
```

and

```
last_data_ms
```

to decide whether the Sender has stopped transmitting.

The timeout itself appears to work correctly.

The problem appears to be the elapsed-time calculation.

---

## Suggested fix

If the code currently resembles

```cpp
uint64_t elapsed = now_ms - last_data_ms;

if (elapsed > TB_SENDER_IDLE_TIMEOUT_MS)
{
    close_client();
}
```

then an underflow can produce

```
18446744073709551564
```

instead of

```
-52
```

One possible fix is

```cpp
if (now_ms >= last_data_ms)
{
    uint64_t elapsed = now_ms - last_data_ms;

    if (elapsed > TB_SENDER_IDLE_TIMEOUT_MS)
    {
        close_client();
    }
}
```

Even better would be

- always using CLOCK_MONOTONIC
- ensuring both timestamps originate from the same clock
- making last_data_ms updates atomic if multiple threads are involved

---

## Why I think this is not a network problem

Immediately before disconnect

- video continues decoding
- Receiver continues rendering
- frame rate remains approximately 30 fps

Only after the false timeout does the Receiver close the socket.

Therefore

```
Receiver
    ↓
close(socket)
    ↓
Sender receives "Connection reset by peer"
    ↓
Virtual display removed
```

rather than the other way around.

---

## HEVC decoder warnings

Immediately after connecting I also see

```
Could not find ref with POC
Error constructing the frame RPS
Skipping invalid undecodable NALU
```

These disappear after the first keyframes and the stream becomes stable.

The disconnect occurs much later while decoding remains stable at ~30 fps.

Therefore I do not believe the decoder warnings are directly related to this bug.

---

## Attached

Receiver stderr log showing

```
no data from sender for 18446744073709551564 ms
```

which appears to be the smoking gun.

[TargetBridgeReceiver-20260806-132921.log](https://github.com/user-attachments/files/30775004/TargetBridgeReceiver-20260806-132921.log)

Betafer:
## Summary

Fix the unsigned idle-watchdog underflow reported in #156, where a short timestamp reversal is interpreted as nearly UINT64_MAX milliseconds and disconnects a healthy session.

The elapsed-time calculation is moved into a small testable helper. When the current monotonic value is lower than the saved activity timestamp, the watchdog rebases the timestamp instead of expiring the connection.

The receive path also refreshes last activity from a new monotonic reading after data is drained, so both values used by the watchdog come from the same clock source.

## Scope

- isolated watchdog helper and tests
- no protocol or timeout changes
- normal ten-second stale-session cleanup is preserved

## Verification

- 12 focused watchdog checks cover normal, exact-timeout and 52 ms backward-jump cases
- Receiver make test: 654 total checks; all pass

Closes #156.

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/Makefile
+++ b/TargetBridge-Receiver/TBReceiverC/Makefile
@@ -34,7 +34,7 @@ ifeq ($(UNAME_S),Darwin)
 LDFLAGS  += -framework ApplicationServices -framework AppKit -framework CoreFoundation -framework CoreGraphics -framework CoreText -framework VideoToolbox -framework CoreMedia -framework CoreVideo -framework IOSurface -framework CoreServices -framework CoreAudio -framework SystemConfiguration
 endif
 
-C_SRC = src/main.c src/net.c src/decoder.c src/display.c src/input_queue.c src/receiver_profile.c src/tb_i18n.c
+C_SRC = src/main.c src/net.c src/decoder.c src/display.c src/input_queue.c src/idle_watchdog.c src/receiver_profile.c src/tb_i18n.c
 OBJC_SRC = src/tb_gesture_bridge.m src/tb_display_tweaks.m
 OBJ = $(C_SRC:.c=.o) $(OBJC_SRC:.m=.o)
 BIN = tbreceiver
@@ -51,7 +51,7 @@ $(BIN): $(OBJ)
 	$(CC) $(CFLAGS) -fobjc-arc -c -o $@ $<
 
 clean:
-	rm -f $(OBJ) $(BIN) $(TEST_BIN) $(TEST_INPUT_QUEUE_BIN) $(TEST_RECEIVER_PROFILE_BIN)
+	rm -f $(OBJ) $(BIN) $(TEST_BIN) $(TEST_INPUT_QUEUE_BIN) $(TEST_IDLE_WATCHDOG_BIN) $(TEST_RECEIVER_PROFILE_BIN)
 
 # Unit tests for the packet parser. net.c is pure POSIX, so this needs no
 # ffmpeg/SDL/pkgconf — it runs anywhere, including CI, with no hardware.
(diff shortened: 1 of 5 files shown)


swellweb:

PR #180, opened 2026-08-21
build: support reproducible app bundles

Betafer:
## Summary

- allow a fixed Receiver version and build number through `APP_VERSION` / `APP_BUILD`
- allow a fixed Sender build number through `TB_SENDER_BUILD`
- let the Sender packaging script build either Debug or Release through `CONFIGURATION`
- keep the generated Xcode project and its source `project.yml` aligned

## Why

The setup assistant and release packaging need deterministic app identities. Timestamp-only builds make it difficult to verify that the tested bundle is the one being installed, while a Debug-only packaging path cannot validate a release candidate.

## Validation

- shell syntax checks pass for both packaging scripts
- Receiver bundle built with version `3.5.1` and build `20260821000101`; both values were verified in its Info.plist and executable
- Sender Release bundle built with build `20260821000102`; the value was verified in the executable
- Sender output is a signed universal Mach-O bundle for arm64 and x86_64

This PR changes only four build-definition lines and their generated-project counterpart.


swellweb:

PR #186, opened 2026-08-21
fix(receiver): keep login-item alerts behind fullscreen

Betafer:
## Context

This is a small follow-up to #162. The existing monitor shield uses the same public window level as Notification Center, so a login-item banner created later can still be ordered above the Receiver. On the 2017 21.5-inch iMac, SDL/macOS could also reset the Receiver from the requested shield level back to level 0 after the delayed fullscreen Space transition.

## Changes

- use the first window level above `kCGStatusWindowLevelKey`
- stay below public pop-up-menu and screen-saver/lock levels
- order the Receiver forward only when correcting a changed level
- react to AppKit fullscreen, screen and Retina-backing change notifications
- perform only three bounded post-transition checks (250 ms, 1 s and 2.5 s)
- remove the observers and restore `NSNormalWindowLevel` on Stop/disconnect

## Safety

The patch does not use private APIs or the screen-saver level. Pop-up menus and the lock screen remain above the Receiver. There is no recurring polling: after 2.5 seconds no delayed check remains, and event observers are installed only for an active monitor session.

## Validation

- `make test` and full Receiver build pass on the `maint-3.5` base
- full integrated suite passes: 752 assertions/checks plus recovery-helper checks
- native Metal shader/pipeline test passes on the iMac's AMD Radeon Pro 560
- focused AppKit lifecycle harness verifies: no recurring polling, event-driven recovery, Stop restore and stale-retry cancellation
- built and code-signature-verified as x86_64 on a 2017 21.5-inch 4K iMac running macOS 13.7.8
- real sender session accepted after restart with its direct-link TCP connection established
- after all bounded checks had expired, WindowServer confirmed both Receiver surfaces persisted at layer 26; the previous one-shot implementation had fallen back to layer 0

### DIFF
--- a/TargetBridge-Receiver/TBReceiverC/src/tb_gesture_bridge.m
+++ b/TargetBridge-Receiver/TBReceiverC/src/tb_gesture_bridge.m
@@ -49,20 +49,96 @@ int tb_receiver_content_display_pixels(uint32_t *width, uint32_t *height) {
 }
 
 static BOOL g_monitor_shield_active = NO;
+static uint64_t g_monitor_shield_generation = 0;
+static NSMutableArray *g_monitor_shield_observers = nil;
+
+static NSWindowLevel tb_monitor_shield_level(void) {
+    NSWindowLevel status_level =
+        (NSWindowLevel)CGWindowLevelForKey(kCGStatusWindowLevelKey);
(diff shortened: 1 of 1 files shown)


swellweb:

